PR #47, opened 2018-12-07
[0018] The canonical flatbuffers proposal

doitian:
Preview: \[ [English](https://github.com/doitian/rfcs/blob/cfb-encoding/rfcs/0000-cfb-encoding/0000-cfb-encoding.md) \]

- [x] Fork https://github.com/dvidelabs/flatcc/blob/master/doc/binary-format.md and provide a locked version of FlatBuffers
- [x] Explain why and how to use nested buffers
- [ ] Benchmark
- [ ] Use RFC # 0018


rfcs/0000-cfb-encoding/0000-cfb-encoding.md
```diff
@@ -0,0 +1,98 @@
...
+
+## The Algorithm
+
+The buffer consists of 3 portions:
+
+- root table pointer
+- vtable portion
+- data portion
+
+Root table pointer is a 4 bytes little endian unsigned integer, which equals the start position of the data portion.
```
xxuejie: Does this mean that the serialized data cannot exceed 4GB? While this won't be a problem now, note that in CKB one cell could grow to 4GB by itself, that will make the enclosing transaction or block bigger than 4GB. Maybe it's time to take this opportunity and bump this to 8GB for future proof reason?

doitian: No, it just means the vtable portion is limited to less than 4G.

The problem is that the Vector and String length is encoded with 4 bytes. A simple solution is using a list of `[ubyte]` to store data.

doitian:
Ready for review

rfcs/0000-cfb-encoding/0000-cfb-encoding.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Data alignment
+
+The table's starting position `t` must be aligned, such that `t` is multiple of 4, and `t + 4` is the multiple of the most significant alignment of the present field in the table. All the fields in the table must be aligned as well.
+
+The vtable's starting position `t` must be a multiple of 2.
+
+The string's starting position `t` must be a multiple of 4.
+
+The vector's starting position `t` must be a multiple of 4, and `t + 4` is the multiple of the item alignment. **Specially**, if a vector of `ubyte` is marked as a nested buffer, `t + 4` must be multiple of 8.
```
janx:
```suggestion
The vector's starting position `t` must be a multiple of 4, and `t + 4` must be the multiple of the item alignment. **Specially**, if a vector of `ubyte` is marked as a nested buffer, `t + 4` must be multiple of 8.
```

janx: > if a vector of `ubyte` is marked as a nested buffer, `t + 4` must be multiple of 8.

What's the reason? Can you give an example?

doitian: See here https://github.com/dvidelabs/flatcc/blob/master/doc/binary-format.md#nested-flatbuffers

I think the rule is wrong here, since the nested buffer may contain a field that has a larger alignment requirement. It should be a configurable FlatBuffer user attribute.

We prefer nested buffers for tx and block header, which allows us to partially read and hash the block of memory directly. Otherwise, because of vtable sharing, the memory is not guaranteed to be consecutive.

rfcs/0000-cfb-encoding/0000-cfb-encoding.md
```diff
@@ -0,0 +1,138 @@
...
+
+CFB (Canonical FlatBuffers) is a restricted variant of FlatBuffers for producing unequivocal transfer syntax.
+
+As described in [FlatBuffers Internals][flatbuffers-internals]:
+
+[flatbuffers-internals]: https://google.github.io/flatbuffers/flatbuffers_internals.html
+
+> On purpose, the format of FlatBuffers leaves a lot of details about where exactly things live in memory undefined, e.g. fields in a table can have any order, and objects to some extent can be stored in many orders. This is because the format doesn't need this information to be efficient, and it leaves room for optimization and extension (for example, fields can be packed in a way that is most compact). Instead, the format is defined in terms of offsets and adjacency only. This may mean two different implementations may produce different binaries given the same input values, and this is perfectly valid.
+
+CFB eliminates all the undefined encodings. For a given data structure, the encoded binary is determined.
```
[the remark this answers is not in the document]
doitian: We can lock a version of https://github.com/dvidelabs/flatcc/blob/master/doc/binary-format.md

doitian:
> I'd like to see some benchmarks comparing CFB with Bitcoin block/tx serialization and RLP in terms of time/space efficiency in this PR.

CFB is believed to have the same performance of FlatBuffers, since it is FlatBuffers.

@quake has already done the comparison in https://github.com/nervosnetwork/serde_bench

However, I'll keep in mind to put an eye on the benchmark.

rfcs/0000-cfb-encoding/0000-cfb-encoding.md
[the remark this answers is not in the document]
janx: that would be good.

janx:
Is this PR ready to merge?

doitian:
I'm sorry but it seems that @github is not going to fix the 500 issue. The team could not open this PR in desktop browser, it is inconvenient to discuss in this PR.  I will recreate a new PR.

PR #57, opened 2018-12-21
add rfc for mmr

kilb:


rfcs/0009-output-commitments/0009-output-commitments.md
```diff
@@ -0,0 +1,107 @@
...
+table MMRProof {
+    output_proofs: [CBMTProof],
+    nodes: [H256],
+    left_peaks: [H256],
+}
+```
+
+`CBMTProof` is defined in RFC#0006, it is used for proving that certain outputs inner a transaction is still unspent. The proofs in `output_proofs` are in **descending breadth order** by the node it proved for. The `nodes` is a list of siblings of the nodes along the path that form leaves to peaks, excluding the nodes already in the path, they are also in **descending breadth order**. The `left_peaks` is a list of peaks that can't be calculated directly. The right most peaks should be combined. Also, the peaks are in **descending breadth order**.
+
+If we want to prove the gray nodes in below figure are unspent, the `nodes` in MMRProof would be `[Node4, Node1]` and the `left_peaks` would be `[Node8]`.
```
janx: I suppose CBMTProof only contains CBMT root, how can you derive `H(root, tx_hash, CBMT_Size)` for Node2 and Node5 using MMRProof only?

rfcs/0009-output-commitments/0009-output-commitments.md
```diff
@@ -0,0 +1,107 @@
...
+
+![](images/05.jpg "Pruned MMR with 5 transactions")
+
+### Delayed Commitments
+
+A meta is added into the block header called outputs_root, it is the root of MMR.
+
+When a fork happen, it may cost a lot of work on MMR's rollback. In order to avoid this, we delay the commitments **m** blocks, which means the block **i** commits to the TXO set state as of block **i-m**, in other words the outputs_root in block **i** is the root of MMR up until block **i-m**. So the STXOs and UTXOs of latest **m** blocks should be cached.
+
+### Proof
```
doitian: It is lack of the algorithm to verify the proof and the reason why it cannot be forged.

rfcs/0009-output-commitments/0009-output-commitments.md
```diff
@@ -0,0 +1,107 @@
...
+
+If we want to prove the gray nodes in below figure are unspent, the `nodes` in MMRProof would be `[Node4, Node1]` and the `left_peaks` would be `[Node8]`.
+
+![](images/06.jpg "MMR with 5 transactions")
+
+The proofs are attached to the block body with outputs, this can be done by miner or node server. The structure is:
+
+```
+table OutputWithIndex {
+    output: OutPut,
```
doitian: Should not be `OutPut`. And what is an Output? If it is the entire transaction output, explain it, use the name in current schema file, and link to the latest schema file.

doitian:
I think the section `Delayed Commitments` and `Proof` are lack of too many details.

The RFC also does not talk about the persistant recommendation. Should we persist the MMR and the cache for delayed commitments?

And how different roles in the network can ultilize the MMR feature? Any use cases that we can get benefit from but cannot from the original AVL solution.

Is the pruning rule a consensus or nodes can set the rules themselves?

rfcs/0009-output-commitments/0009-output-commitments.md
```diff
@@ -0,0 +1,107 @@
...
+
+Merkle Mountain Range(MMR) is first introduced by [Peter Todd][1], it is "a type of deterministic, indexable, insertion ordered merkle tree, which allows new items to be cheaply appended to the tree with minimal storage requirements". "Once an output is added to the TXO MMR it is never removed; if an output is spent its status is updated in place." "Both the state of a specific item in the MMR, as well the validity of changes to items in the MMR, can be proven with logn sized proofs consisting of a merkle path to the tip of the tree."
+
+## TXO Commitments using MMR
+
+TXO Commitments using MMR is posed by [Perter Todd][2], this RFC is mostly same with Perter's design. But there are still some difference in detail.
+
+### Structure
+
+The key of output consist of a transaction hash and an index, so we can't organize the output into a MMR directly. So the outputs inner a transaction are built together into a CBMT(RFC#0006), then these CBMTs are organized into a MMR. So every leaf of MMR will have a corresponding CBMT, and a leaf is hash of CBMT's root, transaction's hash and CBMT's size.
```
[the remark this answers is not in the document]
kilb: Because CKB use OutPoint to specify an output, which is consisted of a transaction hash and an index.

rfcs/0009-output-commitments/0009-output-commitments.md
```diff
@@ -0,0 +1,107 @@
...
+
+Specially, the root of an empty MMR is `H256::zero`.
+
+### Spending
+
+If we want spend an output, we just mark the corresponding CBMT node as `H256::zero` and update the CMBT and MMR.
+
+### Pruning
+
+The MMR can be pruned if there are more than **n** transactions(all the outputs in one transaction is combined into a leaf node) in it. The old transaction is always pruned first. In order to update MMR correctly, if a node is still in the MMR, its sibling should not be pruned.
```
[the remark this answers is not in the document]
kilb: Because it is easy for the implement.

rfcs/0009-output-commitments/0009-output-commitments.md
kilb: We can get CBMT_size from CBMTPRoof and tx_hash from OutPoint.

doitian:
We decided to hold the further development and integration of MMR until we see a sufficient advantage to switch.

PR #60, opened 2019-01-04
[0014] doc: Add cycle limit RFC

xxuejie:


doitian:
assigned 0014

doitian:
lgtm, wait for another pair of eyes.

rfcs/0014-vm-cycle-limits/0014-vm-cycle-limits.md
```diff
@@ -0,0 +1,111 @@
...
+| ECALL       | 0      |
+| EBREAK      | 0      |
+
+### Syscall Cycles
+
+Each syscall in CKB has different rules for consuming cycles:
+
+#### Load TX
+
+*Load TX* syscall first consumes 10 initial cycles, it then measures the size of the serialized transaction data: for every single byte in the data, it consumes 10 more cycles.
```
janx: Where do the numbers (10 initial and 10 for each byte) come from?

rfcs/0014-vm-cycle-limits/0014-vm-cycle-limits.md
```diff
@@ -0,0 +1,111 @@
...
+
+As mentioned above, a new scalar `max_block_cycles` field is added to chain spec as a consensus rule, it puts a hard limit on how many cycles a block's scripts can consume. No block can consume cycles larger than `max_block_cycles`.
+
+Note there's no limit on the cycles for an individual transaction or a script. As long as the whole block consumes cycles less than `max_block_cycles`, a transaction or a script in that block are free to consume how many cycles they want.
+
+## Cycle Measures
+
+Here we will specify the cycles needed by each CKB VM instructions or syscalls. Note right now in the RFC, we define hard rules for each instruction or syscall here, in future this might be moved into consensus rules so we can change them more easily.
+
+### Instruction Cycles
```
janx: How are these instruction cycles determined?

rfcs/0014-vm-cycle-limits/0014-vm-cycle-limits.md
```diff
@@ -0,0 +1,111 @@
...
+
+*Load TX* syscall first consumes 10 initial cycles, it then measures the size of the serialized transaction data: for every single byte in the data, it consumes 10 more cycles.
+
+Note that even though the script only requires part of the serialized TX data, the syscall still charges based on the full serialized data size.
+
+#### Load Cell
+
+*Load Cell* syscall first consumes 100 initial cycles, it then measures the size of the serialized cell structure data: for every single byte in the serialized data, it consumes 100 more cycles.
+
+Notice the charged cycles here is 10 times the cycles charged in `Load Cell By Field` syscall, this is because we are discouraging the use of this syscall. One should only use this if they really need the full serialized Cell structure.
```
janx: Why do we discourage this call?

rfcs/0014-vm-cycle-limits/0014-vm-cycle-limits.md
```diff
@@ -0,0 +1,111 @@
...
+
+Note that even though the script only requires part of the serialized Cell structure data, the syscall still charges based on the full serialized data size.
+
+#### Load Cell By Field
+
+*Load Cell By Field* syscall first consumes 10 initial cycles, it then measures the size of the serialized data from the specified field: for every single byte in the serialized data, it consumes 10 more cycles.
+
+Note that even though the script only requires part of the serialized data, the syscall still charges based on the full serialized data size.
```
janx: So the syscall charges based on serialized cell data size, instead of serialized field data size?

rfcs/0014-vm-cycle-limits/0014-vm-cycle-limits.md
```diff
@@ -0,0 +1,111 @@
...
+
+#### Load Input By Field
+
+*Load Input By Field* syscall first consumes 10 initial cycles, it then measures the size of the serialized data from the specified field: for every single byte in the serialized data, it consumes 10 more cycles.
+
+Note that even though the script only requires part of the serialized data, the syscall still charges based on the full serialized data size.
+
+#### Debug
```
janx: Will `log` be a more appropriate name than `debug`, as the syscall is used to print logs? It feels a bit weird to allow `debug` instruction in mainnet's vm.

rfcs/0014-vm-cycle-limits/0014-vm-cycle-limits.md
xxuejie: This is just made-up number, the idea is syscalls should be relatively more expensive than normal instructions, but not too much (since syscalls are charged by data size).

The initial 10 comes from the fact that we want each syscall to at least consume some cycles since certain syscalls might return `ITEM_MISSING`, which means no data is available.

rfcs/0014-vm-cycle-limits/0014-vm-cycle-limits.md
xxuejie: This is roughly based on SiFive FU540's spec at [here](https://sifive.cdn.prismic.io/sifive%2F834354f0-08e6-423c-bf1f-0cb58ef14061_fu540-c000-v1.0.pdf), there're several basic principles:

* Branches should be more expensive than normal instructions
* Memory accesses should be more expensive than normal instructions, but since we are using 64-bit system, accessing 64-bit value should take less time than non 64-bit value
* Multiplication and divisions should be much more expensive than normal instructions

rfcs/0014-vm-cycle-limits/0014-vm-cycle-limits.md
xxuejie: So at CKB implementation side, what we have is individual fields in `CellOutput` structure, so accessing individual fields is cheap and don't need much preprocessing in most cases. However for this `Load Cell` syscall, we need to re-assemble the full CellOutput binary from Rust structure, if cell data is relatively big, this could mean a lot of memory allocations and copying, hence we are putting much bigger cycle factor to discourage this call.

rfcs/0014-vm-cycle-limits/0014-vm-cycle-limits.md
xxuejie: There's a little confusion, so this is indeed charged based on serialized field data size.

rfcs/0014-vm-cycle-limits/0014-vm-cycle-limits.md
xxuejie: Yes that's probably a good change, but maybe we should solve that in a different PR, since `VM Syscalls` RFC already use `debug` as syscall name?

rfcs/0014-vm-cycle-limits/0014-vm-cycle-limits.md
janx: I feel that syscall cycles should not be affected by implementation details.

rfcs/0014-vm-cycle-limits/0014-vm-cycle-limits.md
janx: sounds good.

PR #62, opened 2019-01-07
using more reasonable proof structure

kilb:
So index will transparent to users.

### DIFF
--- a/rfcs/0006-merkle-tree/0006-merkle-tree.md
+++ b/rfcs/0006-merkle-tree/0006-merkle-tree.md
@@ -57,7 +57,7 @@ Suppose a CBMT with *n* items, the size of the array would be *2n-1*, the index 
 
 ## Merkle Proof
 
-Merkle Proof can provide a proof for existence of one or more items. Only sibling of the nodes along the path that form leaves to root, excluding the nodes already in the path, should be included in the proof. We also specify that ***the nodes in the proof is presented in descending order***(with this, algorithms of proof's generation and verification could be much simple). The indexes of items that need to prove is also included in the proof, and ***presented in ascending order by corresponding hash***. For example, if we want to show that `[T1, T4]` is in the list of 6 items above, only nodes `[T5, T0, B3]` and indexes `[9, 6]` should be included in the proof.
+Merkle Proof can provide a proof for existence of one or more items. Only sibling of the nodes along the path that form leaves to root, excluding the nodes already in the path, should be included in the proof. We also specify that ***the nodes in the proof is presented in descending order***(with this, algorithms of proof's generation and verification could be much simple). Indexes of item that need to prove are essential to complete the root calculation, since the index is not the inner feature of item, so the indexes are also included in the proof, and in order to get the correct correspondence, we specify that the indexes are ***presented in ascending order by corresponding hash***. For example, if we want to show that `[T1, T4]` is in the list of 6 items above, only nodes `[T5, T0, B3]` and indexes `[9, 6]` should be included in the proof.
 
 ### Proof Structure
 
--- a/rfcs/0006-merkle-tree/0006-merkle-tree.md
+++ b/rfcs/0006-merkle-tree/0006-merkle-tree.md
@@ -57,16 +57,16 @@ Suppose a CBMT with *n* items, the size of the array would be *2n-1*, the index 
 
 ## Merkle Proof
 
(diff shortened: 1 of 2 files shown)


rfcs/0006-merkle-tree/0006-merkle-tree.md
```diff
@@ -57,16 +57,16 @@ Suppose a CBMT with *n* items, the size of the array would be *2n-1*, the index
...
 
-Merkle Proof can provide a proof for existence of one or more items. Only sibling of the nodes along the path that form leaves to root, excluding the nodes already in the path, should be included in the proof. We also specify that ***the nodes in the proof is presented in descending order***(with this, algorithms of proof's generation and verification could be much simple). For example, if we want to show that `[T1, T4]` is in the list of 6 items above, only nodes `[T5, T0, B3]` should be included in the proof.
+Merkle Proof can provide a proof for existence of one or more items. Only sibling of the nodes along the path that form leaves to root, excluding the nodes already in the path, should be included in the proof. We also specify that ***the nodes in the proof is presented in descending order***(with this, algorithms of proof's generation and verification could be much simple). The indexes of items that need to prove is also included in the proof, and ***presented in ascending order by corresponding hash***. For example, if we want to show that `[T1, T4]` is in the list of 6 items above, only nodes `[T5, T0, B3]` and indexes `[9, 6]` should be included in the proof.
```
janx: Why is it more reasonable to include item indices in proof?

kilb: Because the tree index will be transparent to the caller, and the caller doesn't need to care about the tree index.

janx: Can you be more concrete and give an example? I need to understand the problem and context first to understand the change.

kilb: In the former design, the caller needs to know the index of each item, but in some scenarios, the caller may have no idea about the index, so the caller may need extra interaction to query the index. It will make the proof verification complicate and interactive.  Ideally, the caller should verify the items without any extra information, beyond the items he has and proof he got.

janx: What's the scenario you refer to? Is it worth to add the scenario description and the reason of including index in proof into rfc?

kilb: In the scenario that index is not known, for example, in most cases, the SPV client got a transaction, since the index is not the feature of transactions, the client has no idea about the index.  So the client has to query the index. But the index is the feature of the Merkle tree, it should be transparent to the client.

rfcs/0006-merkle-tree/0006-merkle-tree.md
```diff
@@ -57,16 +57,16 @@ Suppose a CBMT with *n* items, the size of the array would be *2n-1*, the index
...
 
-### Proof Sturct
+### Proof Structure
 
 The schema of proof struct is:
 
 ```
 table Proof {
-  // size of items in the tree
-  size: uint32;
```
quake: The attacker can create a fake item which still passes the proof verification if we omit item size here. e.g.
A merkle tree with 4 items:
```
                B0
        ┌───────┴───────┐
        B1              B3
    ┌───┴───┐       ┌───┴───┐
    T0      T1      T2      T3
```

A fake item with brute forced hash can be created like this diagram:
```
                B0
        ┌───────┴───────┐
        B1              B3
    ┌───┴───┐       ┌───┴───┐
    T0      T1      T2      T3
                  ┌─┴─┐
                 T4 Brute-forced-hash
```

rfcs/0006-merkle-tree/0006-merkle-tree.md
```diff
@@ -102,24 +103,25 @@ Proof gen_proof(Hash tree[], int indexes[]) {
...
-  }
-  
+bool validate_proof(Proof proof, Hash root, Item items[]) {  
   Queue queue;
-  desending_sort_by_item_index(items);
+  ascending_sort_by_item_hash(items);
   
-  for item in items {
-    queue.push_back((item.hash(), item.index() + Proof.size - 1));
+  for (index,item) in (proof.indexes, items) {
```
quake: How to make sure that client knows the items order is same as proof indexes order?

kilb: We sort items by items' hash, the indexes are also in the same order.

doitian:
It looks reasonable for me. The filtered block can also just return the proof and the included transactions.

PR #66, opened 2019-01-20
🐞[HOLD] Propose a CKB address format

CipherWang:
This format convention only applies to application level.


janx:
1. Follow Bitcoin's rule exactly and use the same prefix makes it hard to distinguish a normal bitcoin address and ckb address, which may cause confusion. I think we need more discussions here.
2. It should be `Address` instead of `Account`.

CipherWang:
> 1. Follow Bitcoin's rule exactly and use the same prefix makes it hard to distinguish a normal bitcoin address and ckb address, which may cause confusion. I think we need more discussions here.
The wrapped payload for Bitcoin address is 160 bits, and 256 bits for CKB. If CKB supports other length of cell lock in the future, we could change the prefix hex to distinguish them.

> 2. It should be `Address` instead of `Account`.
np

zhangsoledad:
Bitcoin proposes new checksummed base32 format called `Bech32` and then define Segregated Witness addresses using it. see https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki, it has many advantages over the old base58, if we want to define one for us, I prefer the new `Bech32`.

CipherWang:
cool, I will reconsider it.

CipherWang:
Change the address format proposal to Bech32 format.

ashchan:
> * Follow Bitcoin's rule exactly and use the same prefix makes it hard to distinguish a normal bitcoin address and ckb address, which may cause confusion. I think we need more discussions here.
> * It should be `Address` instead of `Account`.

My 2 cents on the `Address` change. `Account` would too easily confuse people/ourselves with Ethereum's model.

rfcs/0000-address-format/0000-address-format.md
```diff
@@ -0,0 +1,49 @@
...
+
+CKB Address Format is an application level [lock script][script-define] display recommendation. The lock script consists of three key parameters, including version, args, and binary_hash. In the consideration of user experience, it is necessary to wrap the raw data structure into a verifiable and extensible format.
+
+## Solution
+
+CKB Address Format follows [Bitcoin base32 address format (BIP-173)][bip173] rules, which add a version prefix to the payload (such as binary_hash and private key), and wrap them in **Bech32** encoding and a [BCH checksum][bch].
+
+A Bech32 string is at most 90 characters long and consists of the **human-readable part**, the **separator**, and the **data part**.
+
+The human-readable part is "ckb" for CKB mainnet, and "cbt" for the testnet. The separator is always "1". The data part is consist of version prefix and payload. The version part means the address format version, and currently fixed to 0x00. The payload part encodes lock script by three data fields, payload = type | binary_ref | args.
```
janx: `cbt` should be `ckb` according to our previous discussions.

janx: Can we add a section for examples?

ashchan: You mean ‘ckt’ or ‘cbt’?

CipherWang: ckt, np.

doitian:
I'm sorry but it seems that @github is not going to fix the 500 issue. The team could not open this PR in desktop browser, it is inconvenient to discuss in this PR.  Could you please close this and create a new PR instead?

PR #71, opened 2019-02-11
[WIP]proposal: Occupied capacity increment rate

jjyr:

PR #72, opened 2019-02-13
[0015] Occupied capacity increment rate

jjyr:
This PR removed the confused part from #71

### DIFF
--- a/rfcs/0000-occupied-capacity-increment-rate/0000-occupied-capacity-increment-rate.md
+++ b/rfcs/0000-occupied-capacity-increment-rate/0000-occupied-capacity-increment-rate.md
@@ -1,0 +1,57 @@
+---
+Number: 0000
+Category: Standards Track
+Status: Proposal
+Author: Jinyang Jiang
+Organization: Nervos Foundation
+Created: 2019-2-13
+---
+
+# Occupied capacity increment rate
+
+## Abstract
+
+Capacity is one of the core resources in CKB network, itself is just a number representing how much storage resource a user can use, once a user store something in CKB, the occupied part of storage which we called occupied capacity represents the actually used resource in the network.
+
+Occupied capacity affects CKB network because that every full node must cost the same memory or disk to save these content, it incentives miners to be more centralized which does not make any sense for a decentralized network.
+
+So the solution is to limit the increase occupied capacity in each block, to indirectly control the network's total occupied capacity and the increment rate of occupied capacity.
+
+## Specification
+
+Define variable:
+
+* `MAX_INCREMENT_OCCUPIED_CAPACITY` - the max increment occupied capacity of a block, value: 2 MB
+
+Add new verification rule to CKB consensus:
+
+### Increment occupied capacity verification rule
+
+Each transaction spends old outputs and generates new outputs, so it is easy to get increment occupied capacity of a transaction by using the sum of outputs occupied capacity to minus the sum of inputs consumed occupied capacity.
+By accumulating the increment occupied capacity of transactions in a block we can get increment occupied capacity of the block.
+
+The occupied capacity that a block introduced must be less than or equal to `MAX_INCREMENT_OCCUPIED_CAPACITY`.
+
+Pseudo code:
+``` rust
+let mut consumed_occupied_capacity = 0;
+let mut generated_occupied_capacity = 0;
+for tx in block.commit_transactions() {
+  consumed_occupied_capacity += tx.inputs.map(|input| input.previous_output.occupied_capacity()).sum();
+  generated_occupied_capacity += tx.outputs.map(|output| output.occupied_capacity()).sum();
+}
+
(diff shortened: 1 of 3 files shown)


rfcs/0000-occupied-capacity-increment-rate/0000-occupied-capacity-increment-rate.md
```diff
@@ -0,0 +1,57 @@
...
+``` rust
+let mut consumed_occupied_capacity = 0;
+let mut generated_occupied_capacity = 0;
+for tx in block.commit_transactions() {
+  consumed_occupied_capacity += tx.inputs.map(|input| input.previous_output.occupied_capacity()).sum();
+  generated_occupied_capacity += tx.outputs.map(|output| output.occupied_capacity()).sum();
+}
+
+// NOTICE: this value can be negative
+assert!(generated_occupied_capacity - consumed_occupied_capacity <= MAX_INCREMENT_OCCUPIED_CAPACITY);
```
zhangsoledad: `((generated_occupied_capacity - consumed_occupied_capacity).abs())`

rfcs/0000-occupied-capacity-increment-rate/0000-occupied-capacity-increment-rate.md
```diff
@@ -0,0 +1,57 @@
...
+# Occupied capacity increment rate
+
+## Abstract
+
+Capacity is one of the core resources in CKB network, itself is just a number representing how much storage resource a user can use, once a user store something in CKB, the occupied part of storage which we called occupied capacity represents the actually used resource in the network.
+
+Occupied capacity affects CKB network because that every full node must cost the same memory or disk to save these content, it incentives miners to be more centralized which does not make any sense for a decentralized network.
+
+So the solution is to limit the increase occupied capacity in each block, to indirectly control the network's total occupied capacity and the increment rate of occupied capacity.
+
```
zhangsoledad: You have the reason, but what's the target of the limit policy, match the current level of hardware?
Maybe we should supplement this part.

doitian:
Is total size limit another PR?

jjyr:
> Is total size limit another PR?

Block size limit may affect by some consensus details, so better to wait for consensus paper https://github.com/nervosnetwork/rfcs/pull/71#discussion_r256269567

rfcs/0000-occupied-capacity-increment-rate/0000-occupied-capacity-increment-rate.md
jjyr: This part should be introduced in an RFC that describes on-chain governance mechanics.

rfcs/0000-occupied-capacity-increment-rate/0000-occupied-capacity-increment-rate.md
jjyr: We should not care about the decrease of occupied_capacity.

doitian:
Assign RFC# 0015

rfcs/0015-occupied-capacity-increment-rate/0015-occupied-capacity-increment-rate.md
```diff
@@ -0,0 +1,57 @@
...
+Created: 2019-2-13
+---
+
+# Occupied capacity increment rate
+
+## Abstract
+
+Capacity is one of the core resources in CKB network, itself is just a number representing how much storage resource a user can use, once a user store something in CKB, the occupied part of storage which we called occupied capacity represents the actually used resource in the network.
+
+Occupied capacity affects CKB network because that every full node must cost the same memory or disk to save these content, it incentives miners to be more centralized which does not make any sense for a decentralized network.
```
janx: "does not make any sense" is not very descriptive.

rfcs/0015-occupied-capacity-increment-rate/0015-occupied-capacity-increment-rate.md
```diff
@@ -0,0 +1,57 @@
...
+
+So the solution is to limit the increase occupied capacity in each block, to indirectly control the network's total occupied capacity and the increment rate of occupied capacity.
```
janx: Per block increase capacity limit is not 'the solution' to state explosion problem. It's a fine-grained additive to CKB's token economics that bound the state size.

janx:
I reverted the merge because the abstract is misleading. Please adjust the descriptions before merge.

rfcs/0015-occupied-capacity-increment-rate/0015-occupied-capacity-increment-rate.md
janx: It not 'incentives' but forces miners to be more centralized.

janx:
Please also note this limit may not be necessary if the bound imposed by block cycles limit is smaller.

PR #78, opened 2019-03-08
[0015] Cryptoeconomics for the Nervos Common Knowledge Base

knwang:
Everyone, 

This is a RFC(Request for Comments) for the [Cryptoeconomic Design of the Nervos Common Knowledge Base](https://github.com/nervosnetwork/rfcs/blob/0015-crypto/rfcs/0015-ckb-cryptoeconomics/0015-ckb-cryptoeconomics.md). 

Feedback or general comments are welcome.


rfcs/0015-ckb-cryptoeconomics/0015-ckb-cryptoeconomics.md
```diff
@@ -0,0 +1,338 @@
...
+
+
+### Who Compensates the Miners Over the Long Run?
+
+Security and decentralization are two essential properties of a blockchain network, and they come with a high cost that has to be paid to the operators of the network. Bitcoin's current model has network security entirely paid with transaction fees after all the coins are mined. However, the MoE users have very limited time exposure to the network's security risk, therefore won't be willing to pay for it;  the SoV users have prolonged exposure to the network's security risk and are willing to pay for it, but they produce nearly no transactions.
+
+Bitcoin's consensus mechanism incentivize miners to recognize the longest chain as the network's canonical state. Miner's ongoing supply of hashing power doesn't only provide security for the current block, but the immutability of all the blocks before it on the canonical chain. Relying on the SoV users to make one time payments for the ongoing security protection they receive from miners is not sustainable.
```
mingrui: Very interesting point, I will incrementally add my thoughts. Thx for reading!

The theories about machine rules are good but I think the human element is missing. What would the players do outside of the block rewards and transaction fees?

Let’s discuss this in the context of a game and try to come up with a possible Nash equilibrium. Although I am more inclined towards creating an observable environment rather than pure economic modeling. For example ETC surviving the 51% attack goes completely against the fundamental theory of trust on the blockchain.

Setting up the “how to keep the miners running without block rewards” game: 
We need three things: the players, the actions they can take and the equation they try to optimize.

1. The players are the token Whales.

After all tokens are mined, everyone essentially becomes SoV players. MoE players cannot exist without SoV players. If the medium of exchange has no value, no one will be willing to exchange it. Each token may not have a lot of value but it sure does have to hold it’s value during critical times when it is under attack.

For the sake of simplicity, lets consider mainly two groups of SoV token holders: 1. "Whale" who holds an extremely large amount at a high stake & 2. "Hedgehog" who holds a smaller amount as a "hedge" against tradition assets. The Whales will be the major players, the number of whales is somewhat observable through chain analysis, this number is much smaller than the number of Hedgehogs. Whale token holdings can be extrapolated, the blockchain is at least more transparent than an offshore account.

2. The action the players can take is to pay for electricity with tokens directly, off chain.

To keep enough miners running, payments for electricity does NOT have to happen automatically on-chain through transaction fees NOR block rewards. The Whale has an overwhelming incentive to keep tokens valuable and payment for electricity can be done off-chain through direct token sales on the open market. 

3. The equation we try to optimize

The price for a 51% attack can be calculated, margin of safety should be added to the calculation.

Whales are faced with the following game: within the time period of every block, they have to decide collectively (along with the whole community) how much tokens they are willing to give up to keep the miners running. (There is also a wealth transfer effect from whales selling the tokens on the open market to other participants, need further discussion.)

This is “Proof of Stake” in its purest sense. As a whale, how much token are you willing to sell to others to keep your existing tokens safe? 

This game repeats infinitely every block…

Off topic:
The resulting game is similar to the “Push the Button” game in the TV show LOST, everyone benefits from someone pushing the button every 108 minutes. The game has to be done through human interaction instead of being automated. The players within the game effect each others decisions with grave consequences.

knwang: this feels like a more open ended discussion than feedback to the document itself. let's take this to the forum: https://talk.nervos.org/

rfcs/0015-ckb-cryptoeconomics/0015-ckb-cryptoeconomics.md
```diff
@@ -0,0 +1,337 @@
...
+
+Decentralized multi-assets smart contract platforms have to be Store of Assets to be sustainable.
+
+## 5. Decentralization and the Need for Bounded State
+
+Like other long term store of value systems, a Store of Assets platform has to be neutral and free of risks of censorship and confiscation. These are the properties that made gold the world's favorite the store of value for thousands of years. For open, permission-less blockchain networks, censorship resistance comes down to having the broadest consensus scope with a low barrier for consensus and full node participation. Compared to payment networks, running a full node for a smart contract system is more resource intensive. Therefore a Store of Assets platform must take measures to protect the operating cost of full nodes to keep the network sufficiently decentralized.
```
mingrui: I somewhat agree with gold is neutral, but don't agree with gold being free of risks of censorship and confiscation. For example gold was easily confiscated during the Holocaust.

PR #83, opened 2019-03-12
[0017] rfc: transaction valid since

jjyr:


rfcs/0000-tx-valid-since/0000-template.md
```diff
@@ -0,0 +1,45 @@
...
+
+Transaction adds a new `u64` type field `valid_since`, which prevent the transaction to be mined before a certain point, the highest bit of `valid_since` is `type flag`, the other 63 bits represent the actual value `V`, the consensus to validate this field described as follow:
+
+* ignore this validate rule if all bits of `valid_since` is 0.
+* `V` represent a block number when `type flag` is 0, the validation MUST failed if `tip.block_number < V`.
+* `V` represent a block timestamp when `type flag` is 1, the timestamp represented as `V * 512` seconds, the validation MUST failed if `tip.timestamp < V * 512`.
+
+Otherwise, the transaction validate SHOULD continue.
+
+A cell lock script can check the `valid_since` field of a transaction and return invalid when `valid_since` not satisfied condition, to indirectly prevent cell to be spent before a certain block timestamp or a block number.
```
xxuejie: Maybe I misunderstand something, why use lock script to check for `valid_since` one more time? My feeling is that the above consensus rule already does the checking, so what's the value added when using lock script?

rfcs/0000-tx-valid-since/0000-template.md
```diff
@@ -0,0 +1,45 @@
...
+
+# Transaction valid since
+
+## Abstract
+
+This RFC suggests adding a new consensus rule to prevent a cell to be spent before a certain block timestamp or a block number.
+
+## Summary 
+
+Transaction adds a new `u64` type field `valid_since`, which prevent the transaction to be mined before a certain point, the highest bit of `valid_since` is `type flag`, the other 63 bits represent the actual value `V`, the consensus to validate this field described as follow:
```
xxuejie: This is not necessarily a problem but just want to throw a discussion here: if we look at core, we don't (I think?) yet have a field that encode multiple entries now. One example is `reference` vs `binary` in Script, while we can keep both in a single `Vec<u8>`, we opt to put them in 2 fields and use `Option` to wrap them so we know if a value is present there.

I was thinking maybe we can do the same thing here: we can use a separate flag value and 2 more values for block number and timestamp individually. CFB should help us ensure one missing item won't need the extra space. I know this will be more repetitive than a single field here, but what we gain here, is more clarity in the code. Thoughts?

janx: I think 1-bit flag is simple enough and can be clearly express in code. If CFB incurs the same 1 bit cost on separate fields I'm ok with either way, otherwise I prefer 1-bit flag for succinct.

xxuejie: Sounds fair to me

doitian: Flatbuffers has a significant overhead to encode a union.

jjyr: Use `union` looks good, it's a more modern way to represent `valid_since`, `union` needs a `u8` enum type to represent different value, I think it is not a big deal.

Updated, use `union` to represent the `valid_since` field.

janx: If each block is 1s, `1 << 62` will take 146235604338.768 years to run out. I'm not sure why you're uncomfortable with such a long time.

And I don't think `1 << 63` is much better than `1 << 62`.

janx: Succint is a very important goal for messages that will be transmitted on network. The waste will accumulate quickly if we don't pay attention.

rfcs/0000-tx-valid-since/0000-template.md
jjyr: The above consensus rule can prevent tx to be mined before a point, BUT CAN'T prevent an output to be spent before the point. A user who wants to spend an output can always build another tx without valid_since, so check valid_since in lock script is necessary for the time-based fund lock.

rfcs/0000-tx-valid-since/0000-template.md
jjyr: Fair enough, I'll revert to `1-bit` proposal.

jjyr:
@doitian Please assign an RFC number if no more issues.

rfcs/0000-tx-valid-since/0000-template.md
xxuejie: Understood now, thanks

doitian:
> @doitian Please assign an RFC number if no more issues.

You got 0017

doitian:
FYI https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki

rfcs/0017-tx-valid-since/0017-tx-valid-since.md
```diff
@@ -0,0 +1,114 @@
...
+## Abstract
+
+This RFC suggests adding a new consensus rule to prevent a cell to be spent before a certain block timestamp or a block number.
+
+## Summary 
+
+Transaction adds a new `u64` type field `valid_since`, which prevent the transaction to be mined before a certain point, the highest bit of `valid_since` is `type flag`, the other 63 bits represent the actual value `V`, the consensus to validate this field described as follow:
+
+* ignore this validate rule if all bits of `valid_since` is 0.
+* `V` represent a block number when `type flag` is 0, the validation MUST failed if `tip.block_number < V`.
```
doitian:
```suggestion
* `V` represents a block number when `type flag` is 0, the validation MUST fail if `tip.block_number < V`.
```

rfcs/0017-tx-valid-since/0017-tx-valid-since.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Abstract
+
+This RFC suggests adding a new consensus rule to prevent a cell to be spent before a certain block timestamp or a block number.
+
+## Summary 
+
+Transaction adds a new `u64` type field `valid_since`, which prevent the transaction to be mined before a certain point, the highest bit of `valid_since` is `type flag`, the other 63 bits represent the actual value `V`, the consensus to validate this field described as follow:
+
+* ignore this validate rule if all bits of `valid_since` is 0.
```
doitian:
```suggestion
* ignore this validate rule if all bits of `valid_since` are 0.
```

rfcs/0017-tx-valid-since/0017-tx-valid-since.md
```diff
@@ -0,0 +1,114 @@
...
+* `V` represent a block number when `type flag` is 0, the validation MUST failed if `tip.block_number < V`.
+* `V` represent a block timestamp when `type flag` is 1, the timestamp represented as `V * 512` seconds, the validation MUST failed if `tip.timestamp < V * 512`.
```
doitian:
```suggestion
* `V` represents a block timestamp when `type flag` is 1, the timestamp is represented as `V * 512` seconds, the validation MUST fail if `tip.timestamp < V * 512`.
```

rfcs/0017-tx-valid-since/0017-tx-valid-since.md
```diff
@@ -0,0 +1,114 @@
...
+
+Otherwise, the transaction validate SHOULD continue.
```
doitian:
```suggestion
Otherwise, the transaction validation SHOULD continue.
```

rfcs/0017-tx-valid-since/0017-tx-valid-since.md
```diff
@@ -0,0 +1,114 @@
...
+
+# Transaction valid since
+
+## Abstract
+
+This RFC suggests adding a new consensus rule to prevent a cell to be spent before a certain block timestamp or a block number.
+
+## Summary 
+
+Transaction adds a new `u64` type field `valid_since`, which prevent the transaction to be mined before a certain point, the highest bit of `valid_since` is `type flag`, the other 63 bits represent the actual value `V`, the consensus to validate this field described as follow:
```
doitian:
```suggestion
Transaction adds a new `u64` type field `valid_since`, which prevents the transaction to be mined before a certain point, the highest bit of `valid_since` is `type flag`, the other 63 bits represent the actual value `V`, the consensus to validate this field described as follow:
```

PR #93, opened 2019-03-28
rfc0009: new syscall load_block_via_cell

jjyr:


rfcs/0009-vm-syscalls/0009-vm-syscalls.md
```diff
@@ -234,6 +234,41 @@ This syscall would first locate an input in current transaction via `source` and
...
+* `offset`: the exact same `offset` value as used in *Load Transaction* syscall.
+* `index`: an index value denoting the index of cells to read.
+* `source`: a flag denoting the source of cells to locate, possible values include:
+    + 0: current cell, in this case `index` value would be ignored since there's only one current cell.
+    + 1: input cells.
+    + 2: dep cells.
+* `field`: a flag denoting the field of the block to read, possible values include:
+    + 0: `header`
+    + 1: `uncles`
+    + 2: `commit_transactions`
```
xxuejie: Are we gonna serialize all commit transactions in a block? There are 2 problems with this solution:

1. The data might just be too big, just think this contains many transactions, and one transaction contains many cells, each cell might also have a large data part, this is not gonna scale properly.
2. CFB doesn't allow using vectors as root type, so we cannot just return array of transactions here.

jjyr: Since CKB already represent blocks in CFB format internally, we do not need actually to do the serialize. 

Can users use CFB to calculate `offset` and `len`, and load data in need? @ian

If CFB support this feature maybe we can remove `field` argument, user use `offset` and `len` to indicate which fields to load.

xxuejie: It's not just about how we represented and if a serialization step is needed. The other side of the problem, is that we might feed too large data into a VM with only 16MB memory, which requires extremely difficult way to program this.

I suggest we give it a rethink to see if we really need to feed all previous transactions in a block, to a VM script.

jjyr: For now, we can only support to load blocks header, I' ll update PR.

rfcs/0009-vm-syscalls/0009-vm-syscalls.md
```diff
@@ -234,6 +234,38 @@ This syscall would first locate an input in current transaction via `source` and
 
 Specifying an invalid source value here would immediately trigger a VM error, however specifying `output` as the source here would only result in `2` as return value, specifying `current` as source in a *type* script, which doesn't have input, would also result in `2` as return value. Specifying an invalid index value here, would result in `2` as return value, denoting item missing state. Specifying any invalid field will also trigger VM error immediately. Otherwise the syscall would return `0` denoting success state.
 
+### Load Block Via Cell
```
janx: I prefer to add a more fine-grained syscall, e.g. `load_header_via_cell`. Reasons:

1. It's unlikely we would allow script to load history transactions in future, because it means script would have access to state in dead cells;
2. Usually scripts only need certain fields in header (e.g. timestamp) instead of a complete header;

The `field` parameter can be more find-grained in `load_header_via_cell`, e.g.

```
- "0": "all"
- "1" : "prevhash"
...
```

PR #99, opened 2019-04-14
[0018] The canonical flatbuffers encoding proposal

doitian:
Preview: \[ [English](https://github.com/doitian/rfcs/blob/cfb-encoding/rfcs/0000-cfb-encoding/0000-cfb-encoding.md) \]

- [x] Fork https://github.com/dvidelabs/flatcc/blob/master/doc/binary-format.md and provide a locked version of FlatBuffers
- [x] Explain why and how to use nested buffers
- [ ] Benchmark
- [ ] Use RFC # 0018


doitian:
Cancelled

PR #103, opened 2019-04-23
refactor: update contents about capacity (add "CK Shannon")

yangby-cryptape:
### References
- [CKB PR-528: Refactor Capacity](https://github.com/nervosnetwork/ckb/pull/528)

cc @ashchan @classicalliu

### DIFF
--- a/rfcs/0002-ckb/0002-ckb.md
+++ b/rfcs/0002-ckb/0002-ckb.md
@@ -62,7 +62,7 @@ Possible state generation methods include (but are not limited to) the following
 - Local generators on the client: Generators run directly on the client’s devices. Developers can implement the generator in any programming language.
 - Web services: Users may use traditional web services to generate new states. All current web services may work with CKB in this way to gain more trust and liquidity for the generated states. For example, game companies may define in-game items as assets in CKB, the game itself functions as a web service that generates game data, which is then verified and stored in CKB.
 - State channels: Two or more users may use peer-to-peer communication to generate new states.
-- Generation chains: A generation chain is a blockchain that generates new states and stores them in CKB. Generation chains may be permissionless blockchains or permissioned blockchains. In each generation chain, nodes reach consensus in smaller scopes, providing better privacy and performance. 
+- Generation chains: A generation chain is a blockchain that generates new states and stores them in CKB. Generation chains may be permissionless blockchains or permissioned blockchains. In each generation chain, nodes reach consensus in smaller scopes, providing better privacy and performance.
 
 ![Figure 1. Layered Architecture](images/layered-architecture.png)
 *Figure 1. Layered Architecture*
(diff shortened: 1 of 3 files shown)


rfcs/0002-ckb/0002-ckb.md
```diff
@@ -138,14 +138,14 @@ We refer to programs that generate new states and create new cells as Generators
...
 
 A cell is an immutable object, no one can modify it after creation. Every cell can only be used once, it cannot be used as input for two different transactions. Cell ‘updates’ mark previous cells as history and create new cells with the same capacity to replace them. By constructing and sending transactions, users provide new cells with new states in them and invalidate previous cells that store old states atomically. The set of all current (or live) cells represents the latest version of all common knowledge in CKB, and the set of history (or dead) cells represents all historical versions of common knowledge.
 
-CKB allows users to transfer a cell's capacity all at once, or transfer only a fraction of a cell's capacity, which would in turn lead to more cells being created (e.g., a cell with capacity=10 can become two cells with capacity=5).
+CKB allows users to transfer a cell's capacity all at once, or transfer only a fraction of a cell's capacity, which would in turn lead to more cells being created (e.g., a cell which capacity is 9 bytes can become four cells which capacity are 2.25 bytes).
```
doitian: It is a bit weird to name 2.25 bytes

rfcs/0015-ckb-cryptoeconomics/0015-ckb-cryptoeconomics.md
```diff
@@ -115,6 +115,8 @@ The native token for the Nervos CKB is the "Common Knowledge Byte", or "CK Byte"
 
 A cell's occupied capacity could be equal to or less than its specified capacity. For example, for a 1000 byte cell, 4 bytes would be used to specify its own capacity, 64 bytes for the lock script and 128 bytes for storing state. Then the cell's current occupied capacity is 196 bytes, but with room to grow up to 1000 bytes.
 
+The the smallest unit of the native token is "CK Shannon": `1 CK Byte = 100_000_000 CK Shannons`. "CK Shannon" is the indivisible unit.
+"CK Shannon" is designed for the scenes that people want to transfer value less than one "CK Bytes".
```
doitian:
```suggestion
"CK Shannon" is designed for the scenes that people want to transfer value less than one "CK Byte".
```

rfcs/0002-ckb/0002-ckb.md
ashchan:
```suggestion
CKB allows users to transfer a cell's capacity all at once, or transfer only a fraction of a cell's capacity, which would in turn lead to more cells being created (e.g., a cell whose capacity is 9 bytes can become four cells whose capacity is 2.25 bytes each).
```

ashchan: Not 100% about the grammar here but it doesn't feel quite right.

yangby-cryptape: Yep, I feel weird, too!
I just want to show that bytes are divisible.

How about `10 bytes = 2.5 bytes * 4`?
Or I just keep the example as original?

ashchan: Is `1 byte` the minimum unit for (storage) capacity?

yangby-cryptape: Yes, I think I should split the two usages of capacity to two paragraphs.
I should not mix them, let them be a bit of a mess.

janx: I prefer not to introduce `Shannon` to this RFC, as its purpose is to illustrate the general design of CKB and avoid too much details.

rfcs/0019-data-structures/0019-data-structures.md
```diff
@@ -38,7 +38,7 @@ This documents explains all the basic data structures used in CKB.
 
 | Name       | Type       | Description                                                  |
 | :--------- | :--------- | :----------------------------------------------------------- |
-| `capacity` | uint64     | **The size of the cell.** When a new cell is generated (via transaction), one of the verification rule is `capacity ≥ len(capacity)+len(data)+len(type)+len(lock)`. This value also represents the balance of CKB coin, just like the `nValue` field in the Bitcoin's CTxOut. (E.g. Alice owns 100 CKB coins means she can unlock a group of cells that has 100 amount of `capacity` in total.) |
+| `capacity` | uint64     | **The size of the cell (in shannons).** When a new cell is generated (via transaction), one of the verification rule is `capacity_in_bytes > len(capacity) + len(data) + len(type) + len(lock)`. This value also represents the balance of CKB coin, just like the `nValue` field in the Bitcoin's CTxOut. (E.g. Alice owns 100 CKB coins means she can unlock a group of cells that has 100 amount of `bytes` or 10_000_000_000 amount of `shannons` in total.) |
```
janx:
```suggestion
| `capacity` | uint64     | **The size of the cell (in shannons).** When a new cell is generated (via transaction), one of the verification rule is `capacity_in_bytes > len(capacity) + len(data) + len(type) + len(lock)`. This value also represents the balance of CKB coin, just like the `nValue` field in the Bitcoin's CTxOut. (E.g. Alice owns 100 CKB coins means she can unlock a group of cells that has 100 amount of `bytes` (which is 10_000_000_000 amount of `shannons`) in total.) |
```

Issue #104, opened 2019-04-24
Update RFC0019 for variables rename


doitian:
https://github.com/nervosnetwork/ckb/pull/533

janx:
The information in rfc#0019 is no longer accurate, please update.

cc Haichao Zhu, @xxuejie

PR #108, opened 2019-05-05
fix: change `binary_hash` to `code_hash`

Mine77:


doitian:
It's only a part of https://github.com/nervosnetwork/rfcs/issues/104

Mine77:
I see. Thx.

PR #112, opened 2019-05-07
doc: Add syscall to load current script hash

xxuejie:
Ref https://github.com/nervosnetwork/ckb/pull/651

### DIFF
--- a/rfcs/0009-vm-syscalls/0009-vm-syscalls.md
+++ b/rfcs/0009-vm-syscalls/0009-vm-syscalls.md
@@ -168,7 +168,6 @@ The arguments used here are:
 * `offset`: the exact same `offset` value as used in *Load Transaction* syscall.
 * `index`: an index value denoting the index of cells to read.
 * `source`: a flag denoting the source of cells to locate, possible values include:
-    + 0: current cell, in this case `index` value would be ignored since there's only one current cell.
     + 1: input cells.
     + 2: output cells.
     + 3: dep cells.
@@ -199,7 +198,6 @@ The arguments used here are:
 * `offset`: the exact same `offset` value as used in *Load Transaction* syscall.
 * `index`: an index value denoting the index of cells to read.
 * `source`: a flag denoting the source of cells to locate, possible values include:
-    + 0: current cell, in this case `index` value would be ignored since there's only one current cell.
     + 1: input cells.
     + 2: output cells.
     + 3: dep cells.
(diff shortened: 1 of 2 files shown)


rfcs/0009-vm-syscalls/0009-vm-syscalls.md
```diff
@@ -258,6 +255,26 @@ Specifying an invalid source value here would immediately trigger a VM error, ho
...
+```c
+int ckb_load_current_script_hash(void* addr, uint64_t* len, size_t offset)
+{
+  return syscall(2058, addr, len, offset, 0, 0, 0);
+}
+```
+
+The arguments used here are:
+
+* `addr`: a pointer to a buffer in VM memory space denoting where we would load the serialized transaction data.
```
janx: `serialized transaction data`?

rfcs/0009-vm-syscalls/0009-vm-syscalls.md
```diff
@@ -258,6 +255,26 @@ Specifying an invalid source value here would immediately trigger a VM error, ho
 
 NOTE there is one quirk when requesting `args` part in `deps`: since CFB doesn't allow using a vector as the root type, we have to wrap `args` in a `CellInput` table, and provide the `CellInput` table as the CFB root type instead.
 
+### Load Current Script Hash
```
janx: How about remove `current`, leave the name just `load_script_hash`?

rfcs/0009-vm-syscalls/0009-vm-syscalls.md
```diff
@@ -258,6 +255,26 @@ Specifying an invalid source value here would immediately trigger a VM error, ho
...
+{
+  return syscall(2058, addr, len, offset, 0, 0, 0);
+}
+```
+
+The arguments used here are:
+
+* `addr`: a pointer to a buffer in VM memory space denoting where we would load the serialized transaction data.
+* `len`: a pointer to a 64-bit unsigned integer in VM memory space, when calling the syscall, this memory location should store the length of the buffer specified by `addr`, when returning from the syscall, CKB VM would fill in `len` with the actual length of the buffer. We would explain the exact logic below.
+* `offset`: an offset specifying from which offset we should start loading the serialized transaction data.
```
janx: `serialized transaction data`?

rfcs/0009-vm-syscalls/0009-vm-syscalls.md
```diff
@@ -258,6 +255,26 @@ Specifying an invalid source value here would immediately trigger a VM error, ho
...
+
+This syscall would calculate the hash of current cell and copy it to VM memory space. This is the only part a script can load on its own cell.
```
janx:
```suggestion
This syscall would calculate the hash of current running script and copy it to VM memory space. This is the only part a script can load on its own cell.
```

xxuejie:
Cherry picked in #114

PR #113, opened 2019-05-11
Update syscall rfc

janx:


xxuejie:
Working on revising syscalls now, will cherry-pick this together with #112 into a unified PR later.

PR #117, opened 2019-05-14
timestop [draft]

u2:
I found the timestop in the lightning white paper, but have not known why there is no bip about it. Most of the description comes from the lightning white paper, I think the original description is good enough. Any suggestion?

Other discussions about it: https://news.ycombinator.com/item?id=16125185

### DIFF
--- a/rfcs/00xx-timestop/00xx-timestop.md
+++ b/rfcs/00xx-timestop/00xx-timestop.md
@@ -1,0 +1,66 @@
+---
+Number: "00xx"
+Category: Standards Track
+Status: Draft
+Author: Yaning Zhang
+Created: 2019-05-14
+---
+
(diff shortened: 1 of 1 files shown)


u2:
@doitian

janx:
Yes this is a potential solution to layer2 withdraw problem. As a RFC this document lacks design details.

u2:
> Yes this is a potential solution to layer2 withdraw problem. As a RFC this document lacks design details.

Updated.

u2:
I close it firstly and reopen it later.

jjyr:
1.  `since` support both "epoch number" and "block number", how does this work with epoch number?
2. The timestop value is totally decided by miners, I'm doubting this giving too much power to miners.

PR #129, opened 2019-07-19
RFC for Eaglesong (proof-of-work function)

aszepieniec:
This pull request contains a specification for the Eaglesong hash function, as well as attached source code:
 - C reference implementation
 - python reference implementation
 - python code for generating constants

[direct link](https://github.com/nervosnetwork/rfcs/blob/d9cea94bcd4d846d974023e70d5d9025a06df78c/rfcs/0010-eaglesong/0010-eaglesong.md)


0000-eaglesong/hash.c
```diff
@@ -0,0 +1,29 @@
...
+void main( int argc, char ** argv ) {
+    unsigned char input[10000];
+    unsigned char output[32];
+    int c;
+    int i;
+
+    i = 0;
+    while( 1 ) {
+	    c = getchar();
+	    if( c == EOF ) {
```
TheWaWaR: Mixed use of `<tab>` and `<space>`.

janx:
@aszepieniec please use rfc number 0010.

rfcs/0010-eaglesong/0010-eaglesong.md
```diff
@@ -0,0 +1,335 @@
...
+Category: Standards Track
+Status: Proposal
+Author: Alan Szepieniec
+Organization: Nervos Foundation
+Created: 2019-07-18
+---
+
+# Eaglesong (Proof-of-Work Function for Nervos CKB)
+
+This documents specifies the Eaglesong hash function as it is to be used in the context of Nervos CKB proof-of-work.
```
quake:
```suggestion
This document specifies the Eaglesong hash function as it is to be used in the context of Nervos CKB proof-of-work.
```

rfcs/0010-eaglesong/0010-eaglesong.md
```diff
@@ -0,0 +1,335 @@
...
+<a name="Sponge-Construction"></a>
+## Sponge Construction
+
+Eaglesong is a [sponge construction](https://en.wikipedia.org/wiki/Sponge_function). This means that what is specified is a permutation `F` along with three variables:
+
+ - `r`, the rate : a positive integer;
+ - `c`, the capacity : another positive integer;
+ - `d`, the delimiter : a byte fixed to a nonzero value.
+
+The permutation maps `r+c` bits to `r+c` bits. After appending the delimiter byte to variable-length input to the hash function, this input is chunked into pieces `chunk[i]`, each of `r` bits long with thhe last one padded with zeros as necessary. The sponge construction defines a state, which consists of `r+c` bits initialized to zeros. The absorbing phase iterates over all chunks and xors the current chunk into the top `r` bits of the state before applying the permutation. In the squeezing phase, the output buffer is initialized to the empty string. Until this buffer is of the right size, the permutation is applied to the state and the top `r` bits are read out and appended to the output buffer. If the output buffer is too long, it is truncated. The next pseudocode illustrates this operation.
```
quake:
```suggestion
The permutation maps `r+c` bits to `r+c` bits. After appending the delimiter byte to variable-length input to the hash function, this input is chunked into pieces `chunk[i]`, each of `r` bits along with the last one padded with zeros as necessary. The sponge construction defines a state, which consists of `r+c` bits initialized to zeros. The absorbing phase iterates over all chunks and xors the current chunk into the top `r` bits of the state before applying the permutation. In the squeezing phase, the output buffer is initialized to the empty string. Until this buffer is of the right size, the permutation is applied to the state and the top `r` bits are read out and appended to the output buffer. If the output buffer is too long, it is truncated. The next pseudocode illustrates this operation.
```

rfcs/0010-eaglesong/0010-eaglesong.md
```diff
@@ -0,0 +1,335 @@
...
+# Round Function
+
+The Eaglesong permutation F is obtained by iterating a round function f for `N = 43` times. The round function consists of four steps.
+
+The state of the Eaglesong permutation consists of 16 integers.
+
+<a name="Bit-Matrix"></a>
+## Bit Matrix
+
+The state vector is mapped to a new state vector determined via (row-)vector-matrix multiplication, where the coefficients of the matrix are either 0 or 1. Every row of this matrix represents an indications of which state elements to xor together.
```
quake:
```suggestion
The state vector is mapped to a new state vector determined via (row-)vector-matrix multiplication, where the coefficients of the matrix are either 0 or 1. Every row of this matrix represents an indication of which state elements to xor together.
```

rfcs/0010-eaglesong/0010-eaglesong.md
```diff
@@ -0,0 +1,335 @@
...
+        output_vector[i] = 0
+        for k in [0:3] do:
+            output_vector[i] = output_vector[i] xor (state_vector[i] <<< coefficients[i][k])
+    return output_vector
+```
+
+<a name="Injection-of-Constants"></a>
+## Injection of Constants
+
+A given constant is xorred into every component of the state vector. The coefficients differ per component and per round. They are determined from using Keccak to expand the ASCII string
```
quake:
```suggestion
A given constant is xored into every component of the state vector. The coefficients differ per component and per round. They are determined from using Keccak to expand the ASCII string 
```

rfcs/0010-eaglesong/0010-eaglesong.md
aszepieniec: Not sure what the suggested modification is here (line 52).

quake: it looks like github diff function is broken... just typo fix:
```diff
- each of `r` bits long with thhe last one padded with zeros as necessary.
+ each of `r` bits along with the last one padded with zeros as necessary. 
```

aszepieniec: Got it, thanks!

rfcs/0010-eaglesong/eaglesong.c
```diff
@@ -0,0 +1,199 @@
...
+        state[i] = 0;
+    }
+
+    // absorbing
+    for( i = 0 ; i < ((input_length+1)*8+rate-1) / rate ; ++i ) {
+        for( j = 0 ; j < rate/32 ; ++j ) {
+            integer = 0;
+            for( k = 0 ; k < 4 ; ++k ) {
+                if( i*rate/8 + j*4 + k < input_length ) {
+                    integer = (integer << 8) ^ input[j*4 + k];
```
quake: bug? using the wrong index:

```suggestion
                    integer = (integer << 8) ^ input[i*rate/8 + j*4 + k];
```

rfcs/0010-eaglesong/eaglesong.py
```diff
@@ -0,0 +1,183 @@
...
+
+    state = [0 for i in range(0, 16)]
+
+    # absorbing
+    for i in range(0, ((len(input_bytes)+1)*8+rate-1) // rate):
+        for j in range(0, rate//32):
+            integer = 0
+            for k in range(0, 4):
+                if i*rate//8 + j*4 + k < len(input_bytes):
+                    integer = (integer << 8) ^ input_bytes[j*4 + k]
```
quake: bug? using the wrong index:

```suggestion
                    integer = (integer << 8) ^ input_bytes[i*rate//8 + j*4 + k]
```

rfcs/0010-eaglesong/0010-eaglesong.md
```diff
@@ -0,0 +1,335 @@
...
+    return output_vector
+```
+
+<a name="Injection-of-Constants"></a>
+## Injection of Constants
+
+A given constant is xored into every component of the state vector. The coefficients differ per component and per round. They are determined from using Keccak to expand the ASCII string 
+
+```
+The various ways in which the knowledge on which people base their plan is communicated to them is the crucial problem for any theory explaining the economic process, and the problem of what is the best way to utilizing knowledge initially dispersed among all the people is at least one of the main problems of economic policy - or of designing an efficient economic system.
```
rink1969: What's these words means?
It looks like nothing to do with the context.

rfcs/0010-eaglesong/eaglesong.c
```diff
@@ -0,0 +1,199 @@
...
+0x5c9a812e ,  0x6611fd59 ,  0x85e46044 ,  0x1981d885 ,  0x5a4c903f ,  0x43f30d4b ,  0x7d1d601b ,  0xdd3c3391 ,
+0x030ec65e ,  0xc12878cd ,  0x72e795fe ,  0xd0c76abd ,  0x1ec085db ,  0x7cbb61fa ,  0x93e8dd1e ,  0x8582eb06 ,
+0x73563144 ,  0x049d4e7e ,  0x5fd5aefe ,  0x7b842a00 ,  0x75ced665 ,  0xbb32d458 ,  0x4e83bba7 ,  0x8f15151f ,
+0x7795a125 ,  0xf0842455 ,  0x499af99d ,  0x565cc7fa ,  0xa3b1278d ,  0x3f27ce74 ,  0x96ca058e ,  0x8a497443 ,
+0xa6fb8cae ,  0xc115aa21 ,  0x17504923 ,  0xe4932402 ,  0xaea886c2 ,  0x8eb79af5 ,  0xebd5ea6b ,  0xc7980d3b ,
+0x71369315 ,  0x796e6a66 ,  0x3a7ec708 ,  0xb05175c8 ,  0xe02b74e7 ,  0xeb377ad3 ,  0x6c8c1f54 ,  0xb980c374 ,
+0x59aee281 ,  0x449cb799 ,  0xe01f5605 ,  0xed0e085e ,  0xc9a1a3b4 ,  0xaac481b1 ,  0xc935c39c ,  0xb7d8ce7f };
+
+int num_rounds = 43;
+int capacity = 256;
```
rink1969: unused?

rfcs/0010-eaglesong/eaglesong.c
```diff
@@ -0,0 +1,199 @@
...
+                new[j] = new[j] ^ (bit_matrix[k*16 + j] * state[k]);
+            }
+        }
+        for( j = 0 ; j < 16 ; ++j ) {
+            state[j] = new[j];
+        }
+
+        // circulant multiplication
+        for( j = 0 ; j < 16 ; ++j ) {
+            state[j] = state[j]  ^  (state[j] << coefficients[3*j+1]) ^ (state[j] >> (32-coefficients[3*j+1]))  ^  (state[j] << coefficients[3*j+2]) ^ (state[j] >> (32-coefficients[3*j+2]));
```
rink1969: Is this code not equal to RFC?
In RFC the function is:
```
function ApplyCirculantMultiplication( state_vector ):
    for i in [0:16] do:
        output_vector[i] = 0
        for k in [0:3] do:
            output_vector[i] = output_vector[i] xor (state_vector[i] <<< coefficients[i][k])
    return output_vector
```
Here code unrolling the inner loop.
But output_vector[i] should be changed each itertor.
But the c code use the original value.

rink1969: ok, I see.
No problem.

rfcs/0010-eaglesong/eaglesong.c
rink1969: Is the index of input should add i*rate/8?
Now for each loop of i, we fetch same value from input.
Nerver fetch value beyond input[32].

rink1969: test case
# rink @ rink-pc in ~/work/eaglesong [15:31:07] 
$ ./hash
1111111111111111111111111111111111111111111111
76a52594935beb5d00273d9989c0c594da66f45f4e4b22a1907afbcfbaf2f4e2

# rink @ rink-pc in ~/work/eaglesong [15:32:53] 
$ ./hash
1111111111111111111111111111111122222222222222
76a52594935beb5d00273d9989c0c594da66f45f4e4b22a1907afbcfbaf2f4e2

rfcs/0010-eaglesong/eaglesong.c
```diff
@@ -0,0 +1,199 @@
...
+            state[j] = state[j] ^ integer;
+        }
+        EaglesongPermutation(state);
+    }
+
+    // squeezing
+    for( i = 0 ; i < output_length/(rate/8) ; ++i ) {
+        for( j = 0 ; j < rate/32 ; ++j ) {
+            for( k = 0 ; k < 4 ; ++k ) {
+                output[i*rate/8 + j*4 + k] = (state[j] >> (8*k)) & 0xff;
```
rink1969: Is the index of state should add i*(rate/32)?

rink1969:
There are not much detail about padding and output in RFC.

rfcs/0010-eaglesong/eaglesong.c
rink1969: I've got it.

rfcs/0010-eaglesong/eaglesong.py
aszepieniec: You're right, great catch! I guess the utility of having two independent implementations is undermined when you base one off the other ...

rfcs/0010-eaglesong/0010-eaglesong.md
aszepieniec: This quote by the Nobel prize winning Austrian economist FA Hayek captures concisely the Austrian school's perspective on the function of money within society: money and prices is the mechanism by which scarce resources are directed towards those needs most urgently wanted by society. Money, whether provided by a free market or an intervening state, lies at the heart of any sound understanding of society and how it operates -- not because everyone wants it but because of the problem it solves, i.e., the most-wanted allocation of scarce resources. Any theory of economics that fails to characterize the price mechanism as a network for rapid information dispersal, or free market money as a foundation for it, is necessarily incomplete.

This quote is relevant in the broader context of Nervos CKB because it describes exactly the problem landscape it was designed for. On the one hand, Nervos CKB offers a native currency, which enables a price mechanism and can in principle solve the distribution problem outright -- though more accurately it represents one building block among many with various engineering tradeoffs hoping to solve the same problem. On the other hand, the Common Knowledge Base (which is what CKB stands for) provides another type of knowledge, different from the one Hayek describes. In Hayek's perspective, the knowledge that makes society work is distributed across the minds of all individuals involved. In the CKB, certain knowledge is held and agreed upon by all participants. The interplay between Hayekian knowledge and consensus knowledge is the technical key ingredient that makes off-chain scaling solutions such as the Lightning Network work. By optimizing for the distribution of multi-functional; consensus knowledge, and by enabling the distribution of Hayekian knowledge as a result of a free price system, Nervos aims to provide new and better solutions to the economical problems that society faces.

In the specific context of Eaglesong, the purpose of the quote is to make the argument that the constants are indeed random and not chosen specifically to hide a trapdoor that only the designers know. The argument is 90% complete merely by using Keccak: the output is indistinguishable from random and yet uniquely determined by the input. The argument is completed for the remaining 10% by choosing an arbitrary fixed input string that simultaneously makes it obvious that it was not chosen because the other options lead to less usable output constants. For this purpose, what better arbitrary fixed input string is there to seed Keccak with, but the FA Hayek quote that captures Nervos CKB's target application landscape?

rfcs/0010-eaglesong/eaglesong.c
aszepieniec: I do not understand this comment. Care to elaborate?

rfcs/0010-eaglesong/eaglesong.c
aszepieniec: You probably see the resolution but for the sake of those that don't: what is written as two lines in the RFC, namely the loop iterating over k, is written as one line in the C code.

rfcs/0010-eaglesong/eaglesong.c
aszepieniec: Yes, you're right. Great catch! It should be fixed in the most recent commit. For reference:
```
>$ ./hash 
1111111111111111111111111111111111111111111111
d5c1b3ea9c76712bd3da03cc0f330fa5802e25c2f7c7e43dc0fd7db7dbb9f106
>$ ./hash 
1111111111111111111111111111111122222222222222
2e079d6f923344630905faed8efaa8c1b13c501e454e083680903e61da6d18fd
```

rfcs/0010-eaglesong/eaglesong.c
rink1969: I mean the var capacity is unused in this code.

aszepieniec: Correct. It's there for the sake of completeness: a sponge function necessarily comes with two parameters, rate and capacity. In fact, capacity is implicitly used by treating the state as 16 integers of 32 bits, or 512 bits in total. `512 == rate + capacity`

CXdickens:
Both the C reference implementation and the python reference implementation have a common problem, the "absorbing" part of the EaglesongSponge function. The input data can only be absorbed by 32 bytes, and the excess input will be discarded, resulting in the same hash value.
（1）
# ./hash
11111111111111111111111111111111112
bad1ae97ba833fa56a9d1084e94fda88d8ed53485197134289431dddf20899a2
（2）
 ./hash
11111111111111111111111111111111113
bad1ae97ba833fa56a9d1084e94fda88d8ed53485197134289431dddf20899a2

rfcs/0010-eaglesong/0010-eaglesong.md
```diff
@@ -0,0 +1,335 @@
...
+    state = zeros(r+c)
+
+    // absorb
+    for chunk in chunks do:
+        state[0:r] = state[0:r] xor chunk
+        state = F(state)
+
+    // squeeze
+    output = ""
+    while length(output) <= output_length do:
```
doitian: Is it required that the generated output buffer must be larger than the required length?

aszepieniec: Actually, no -- great catch. It's perfectly fine to use `<` instead of `<=` in the while loop's conditional statement. The input-output behavior of the function is the same, but your solution achieves that with less work.

Technically, the sponge construction is defined to have an infinitely long output. Fixed output length derivated functions just truncate this infinitely long string. And of course it doesn't make much sense to run the permutation after you already have all your output bits :)

rfcs/0010-eaglesong/0010-eaglesong.md
```diff
@@ -0,0 +1,335 @@
+---
+Number: 0010
```
jjyr:
```suggestion
Number: "0010"
```

PR #130, opened 2019-07-23
Obsolete 0008 Temporarily

doitian:
Keep this PR to remind that 0008 is obsoleted. We'll replace it with a new serialization method.


janx:
I think it's better to write down the new serialization here instead of render it obsolete, as this rfc is still a proposal.

PR #141, opened 2019-10-26
Modify 0x01 short format to fit multisig lock standard.

CipherWang:


jjyr:
LGTM

rfcs/0021-ckb-address-format/0021-ckb-address-format.md
```diff
@@ -38,13 +38,28 @@ payload = 0x01 | code_hash_index | single_arg
...
 \* The blake160 here means the leading 20 bytes truncation of Blake2b hash result.
-\* The hash160 here means Bitcoin address calculation *double hash* algorithm, which means `hash160(PK) := Ripemd160(SHA256(PK))`.
+
+\*\* The *multisig script hash* is the 20 bytes blake160 hash of multisig script. The multisig script should be assembled in the following format:
+
+```
+S | R | M | N | Pubkey1 | Pubkey2 | ...
+```
+
+Where S/R/M/N are four single byte unsigned integers, ranging from 0 to 255, and Pubkeys are SECP256K1 public keys. S is format version, currently fixed to 0. M/N means the user must provide M of N signatures to unlock the cell. And R means the provided signatures at least match the first R items of the Pubkey list.
```
janx: What's the use of S?

jjyr: In the original design S is a reserved field for future use, occupied 1 byte.

janx: For what kind of use? If there's no explicit use I prefer to remove it. We can use a new `code_hash_index` if we want to add a new multisig format in the future.

jjyr: Not sure yet, it brings some extensibility, 
we may use it to store flags to enable the new features. Or use it as a version to allow compatible changes in the future.

Of cause, use `code_hash_index ` can also achieve the above. what do you think? @CipherWang

jjyr: The code https://github.com/nervosnetwork/ckb-system-scripts/pull/60 is supposed to be freeze, we decide to do not touch the current code, we still keep the `S`.

doitian: There are two kind of pub keys, compressed and uncompressed. It's not clear which one is used here.

CipherWang: I believe it should be uncompressed version, right?

CipherWang: Can you drop a hint here? @jjyr

doitian: Compressed

In the code:

```
PUBKEY_SIZE 33
```

Double check with @jjyr

rfcs/0021-ckb-address-format/0021-ckb-address-format.md
```diff
@@ -38,13 +38,28 @@ payload = 0x01 | code_hash_index | single_arg
...
+|      0x00       | SECP256K1 + blake160 |     Type     |  blake160(PK)*          |
+|      0x01       | SECP256K1 + multisig |     Type     |  multisig script hash** |
 
 \* The blake160 here means the leading 20 bytes truncation of Blake2b hash result.
-\* The hash160 here means Bitcoin address calculation *double hash* algorithm, which means `hash160(PK) := Ripemd160(SHA256(PK))`.
+
+\*\* The *multisig script hash* is the 20 bytes blake160 hash of multisig script. The multisig script should be assembled in the following format:
+
+```
+S | R | M | N | Pubkey1 | Pubkey2 | ...
```
xxuejie: Please note the multisig script is changed to use the blake160 hash of public keys here instead of directly using public keys: https://github.com/nervosnetwork/ckb-system-scripts/pull/61

rfcs/0021-ckb-address-format/0021-ckb-address-format.md
doitian: see https://github.com/nervosnetwork/rfcs/pull/141#pullrequestreview-309693715

rfcs/0021-ckb-address-format/0021-ckb-address-format.md
CipherWang: I will modify the expression.

PR #158, opened 2019-12-09
propose POA testnet

jjyr:
Updates:

* 2019-12-19: update off-chain governance
* 2019-12-18: update block attestation and verification

--------------

The Aggron testnet is use POW. Unfortunately, the average block time can be very slow even up to a few minutes due to mining power join and leave, see the peak from the chart https://explorer.nervos.org/aggron/charts.

To provide a stable testnet for development I propose a POA testnet to replace the current Aggron.

* The block time should be very stable, approximate to 8 seconds.
* The POA testnet supposed to be a long term running testnet, the validators of the testnet should be maintained and governance by the community.
* A minority malicious validators can't halt or censorship on POA testnet.

[view file](https://github.com/jjyr/rfcs/blob/poa-testnet/rfcs/0000-poa-testnet/0000-poa-testnet.md)

### DIFF
--- a/rfcs/0000-poa-testnet/0000-poa-testnet.md
+++ b/rfcs/0000-poa-testnet/0000-poa-testnet.md
@@ -1,0 +1,124 @@
+---
+Number: "0000"
+Category: <TBD>
+Status: <TBD>
+Author: <TBD>
+Organization: <TBD>
+Created: <TBD>
+---
+
+# POA testnet
+
+## Motivation
+
+The original intention of aggron testnet is to provide a development-friendly testnet,
+ allow develops to deploy and test contracts, it's supposed to be similar to mainnet and provide stable services.
+
+However, the mining power on aggron is very unstable[chars](https://explorer.nervos.org/aggron/charts), Obviously for a reasonable miner there is no incitive to mine the testnet coins, and any new miner or mining pool join the testnet will speedup the block produces time temporally, then they leave, the testnet become extremely slow, the avg block time becomes few minutes or even longer.
+
+A POW blockchain only works when the miner has economic incentives. So we propose a new POA consensus testnet to serve contract development purposes.
+
+We expect the new POA testnet to be long term stable, the testnet should be just like the mainnet without using POW. A contract developer should feel no difference when developing on POA testnet and mainnet.
+
+In the purpose we design the POA testnet with the following principles:
+
+* There should be a dynamic group of validators to maintain the POA network, each validator can submit block and vote. With enough votes, the network can allow new validators to join or evict a exists validators.
+* The network cannot be halted or concerned by minority malicious validators, malicious validators should be eventually evicted by majority honest validators.
+* The other part of the network should be the same as the mainnet(unless block header may need some extract context to do verification).
+
+## POA protocol
+
+For easy to describe the protocol, we define the following variables:
+
+* `VALIDATOR_COUNT` - Number of current validators, this value changed due to new validators join or old validator leaves.
+* `ATTEST_INTERVAL` - A validator cannot attest two blocks within `ATTEST_INTERVAL` number. For example, a validator who attest block (6) must wait for at least `ATTEST_INTERVAL` blocks to do next attest: block (6 + `ATTEST_INTERVAL` + 1). Notice when the `VALIDATOR_COUNT` <= `ATTEST_INTERVAL`, the POA testnet will stuck forever due to no validators can attest a new block.
+* `BLOCK_INTERVAL` - the interval of blocks, set to 8 seconds.
+* `VOTE_LIMIT` - The least votes to make a new validator join or to evict an old validator, should be at least `VALIDATOR_COUNT / 2 + 1`.
+
+## attest a new block
+
+let's pre assuming the validator list already exists. Our protocol is simple enough that the change of validator list will not affect our purpose, we can assume the validator list is fixed for now.
+
+Validators use a round-robin style to attest blocks.
+For block `n`, the validator which index equals `n % VALIDATOR_COUNT` suppose to be the attester. However, a validator can fail to produce a block in time due to network congestion or other reasons, in this case, other validators should produce the block `n`.
+
+Validators can use a simple strategy to achieve this:
+
+* If `INDEX == n % VALIDATOR_COUNT`, wait for `BLOCK_INTERVAL` seconds then produce a new block with difficulty set to `2`.
+* If `INDEX != n % VALIDATOR_COUNT`, wait for `BLOCK_INTERVAL + rand(VALIDATOR_COUNT) * 0.5` seconds then produce a new block with difficulty set to `1`.
+* If the validator is the attester of the last block, wait for `ATTEST_INTERVAL` blocks then continue this strategy.
+
+`ATTEST_INTERVAL` is used for preventing malicious validators to censors the POA network. When malicious validators are less than `ATTEST_INTERVAL + 1` that at least an honest validator have the chance to submit votes and evict malicious validators.
+
+`ATTEST_INTERVAL` can be set to `VALIDATOR_COUNT / 2` the honest validators could eventually evict malicious validators unless the half of validators are corrupted.
+
+## validator list
+
+Validator list can be represented as a list of pubkey hash.
+
+``` txt
+pubkey_hash_1 | pubkey_hash_2 | pubkey_hash_3 ...
+```
+
+We can use an `M of N` multisig lock to describe the validator list, the `M` is set to `VOTE_LIMIT`, and `N` is `VALIDATORS_COUNT`.
+
+To change the validator list, a validator can collect enough votes(the signatures) and send a tx to spent the old multisig lock and construct a new multisig lock with new pubkey hashes. A validator node can use `type_id` to track this validator list.
+
+The pre-defined [multisig script](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0021-ckb-address-format/0021-ckb-address-format.md#short-payload-format) is satisfied the requirements, one except is the `multisig script` should be public so other nodes can check new validator list. Validators should make sure the entire `multisig script` is stored to cell data, and only one cell is constructed to represent the validator list.
+
+Notice, the process of collect data is not included in the POA protocol for simplifying, a responsible validator should check the vote tx carefully before signing it.
+
+## POA header
+
+Unfortunately, CKB header is fixed-length, the `nonce` field takes 128 bits, we can't just put 256 bits secp256k1 signature into the header.
+
+For easier implementation, instead change the block header structure, we put an extra POA context concatenated after cellbase transaction's first witness. To verify a header attestation, we need both the header and the cellbase transaction.
+
+The structure of `POAContext`:
+
+``` txt
+table POAContext {
+    signature:              Byte65,
+    transactions_count:     Uint32,
+    raw_transactions_root:  Byte32, // witness transactions root
+    merkle_proof:           Byte32Vec, // merkle proof of cellbase and voting txs
+    voting_txs:             TransactionVec,
+}
+```
+
+The first field `signature` is a secp256k1 signature signed by a validator.
+
+The signature message is supposed to digest the whole block except the signature itself.
+
+Because the signature is put into the cellbase transaction's witness, and the cellbase itself is digested by the `transaction_root` in the header. If we change the cellbase transaction the `transaction_root` and block hash will also be changed.
+
+We use merkle proof to prove the only changes of the header hash is caused by the `signature` field.
+
+* `merkle_proof` - a merkle proof to prove that cellbase transaction belongs to a `witness_transactions_root`.
+* `raw_transactions_root` - raw transactions root of current block.
+* `transactions_count` - number of transactions.
+
+A node does the following steps to verify the POA header:
+
+1. verify merkle proof is correct.
+    1. calculate `witness_transactions_root` from cellbase and merkle proof `merkle_proof.root(cellbase_hash, transactions_count)`
+    2. calculate `transactions_root = merkle_root([witness_transactions_root, raw_transactions_root])`
+    3. check `transactions_root` equals to `header.transactions_root`
+2. zerorize `signature` field of POAContext in cellbase transaction.
+    1. extract `POAContext` from cellbase transaction witness
+    2. set `signature` field to "0x000...0000(32bytes)"
+    3. set `POAContext` back to cellbase transaction witness
+3. calculate signing message from zerorized cellbase transaction.
+    1. calculate `transactions_root` with zerorized cellbase transaction
+    2. replace header's old `transactions_root` then calculate `blake2b_256(header)` as signature message
+4. recovery the pubkey from `signature` then check the pubkey hash is belongs to a validator.
+
+Because we have a dynamic validators group, To keep enough information to verify a header, we also put the voting transactions into `voting_txs` field, a validator can verify header and get newest validator groups without download the whole blocks
+
+## changes on CKB
+
+(TODO)
+
+## references
+
+1. [Rinkeby proposal](https://github.com/ethereum/EIPs/issues/225)
--- a/rfcs/0000-poa-testnet/0000-poa-testnet.md
+++ b/rfcs/0000-poa-testnet/0000-poa-testnet.md
@@ -36,7 +36,7 @@ We define the following variables:
 
 ### attest a new block
 
-Instead of POW mining, a list of validators plays the role of producing blocks in POA testnet. Validators work in a round-robin style; in each block height, there exists a corresponded validator sign the block with its private key to attest to a new block. For convenient, we call the validator who attest a block an **attester** to distinguish it from other validators.
+Instead of POW mining, a list of validators plays the role of producing blocks in POA testnet. Validators work in a round-robin style; in each block height, there exists a corresponded validator that attests to the new block by sign the block with its private key. For convenient, we call the validator who attest a block an **attester** to distinguish it from other validators.
 
 The community chooses the initial validator list of the POA network by some off-chain governance mechanism, which does not cover in this proposal. At this moment, we can ignore the possible updating of the validator list; our protocol is simple enough works for both dynamic and fixed validators.
 
@@ -78,7 +78,11 @@ So we choose to use a cell to represent the POA validator list. The cell contain
 0 | 0 | VOTE_LIMIT | VALIDATOR_COUNT | blake160(Pubkey1) | blake160(Pubkey2) | ...
 ```
 
-We use a cell with `M of N` multisig lock to describe the validator list, the'M` set to `VOTE_LIMIT`, and `N` set to `VALIDATORS_COUNT`.
+We use a cell with `M of N` multisig lock to describe the validator list, the`M` set to `VOTE_LIMIT`, and `N` set to `VALIDATORS_COUNT`, we need an extra parameter to represent `ATTEST_INTERVAL`, so the final data is a 4 bytes little-endian number to represent `ATTEST_INTERVAL` plus a `multisig script`:
+
+``` txt
+ATTEST_INTERVAL(4 bytes little-endian) | 0 | 0 | VOTE_LIMIT | VALIDATOR_COUNT | blake160(Pubkey1) | blake160(Pubkey2) | ...
+```
 
 To change the validator list, anyone that collects enough votes(the signatures) can send a transaction to update the old validator list cell and construct a new validator list cell.
 
@@ -87,6 +91,22 @@ For simplify to track this cell, we assign a `type_id` on it; validators or anyo
 Before sign a vote transaction, validators must make sure the `multisig script`  in the cell is valid and corresponding to the lock, and only one cell constructed to represent the validator list; otherwise, the POA consensus wound broken.
 
 To simplify, the process of collecting votes do not include in the POA protocol.
+
+### off-chain governance
+
+The main intention of this document is to describe the POA testnet protocol. However, to better explain how the POA testnet validator works, we add this section as a suggestion to the off-chain governance, notice this section is more like a suggestion than a specification.
+
+The intention of on-chain governance and dynamic validators design is to decentralize the POA testnet. Instead of the foundation itself, the community should finally maintain the POA testnet validators.
+
+In the initial stage of the POA testnet, there may be just a small group validator that invited by the Nervos foundation. The validators together should governance the testnet and invite more validators from the community. A candidate validator may be a company or organization that influences the community. 
+
+A possible way to elect candidate validators: 
+
+The exists validators deploy a permissionless voting DAPP on the mainnet allows any company and organization can register them as a candidate. 
+Anyone can deposit mainnet coins to vote on the candidates. 
+The POA testnet validators obey the result of the DApp to invites the candidate to become the testnet validator.
+
+Such an election should periodically host.
 
 ## POA header
 
@@ -112,7 +132,7 @@ The first field `signature` is a secp256k1 signature signed by a validator. The 
 
 Since we put the signature into the cellbase transaction's witness, and the cellbase itself is digested by the `transaction_root` in the header. If we directly update the signature, the `transaction_root` and block hash also changed.
 
-To simply solve this issue, an attester needs to zero the signature field of `POAContext`, then re-compute the block hash as signature's message, after signing replace the zero signature with actually signature.
+To solve this issue, an attester needs to zero the signature field of `POAContext`, then re-compute the block hash as signature's message, after signing replace the zero signature with actually signature.
 
 An attester also needs to provide `cellbase transaction` and merkle proof to convince other validators in the [header-first synchronize]. So we also defined the following fields in the `POAContext`:
 
@@ -142,6 +162,15 @@ To solve this issue, we require `POAContext` also provides the update informatio
 So the `voting_txs` field of `POAContext` is introduced. When a block contains voting transactions, the attester must put voting transactions into `POAContext`, and the `merkle_proof` must proof both cellbase transaction and voting transactions are leaves of the `witness_transactions_root`.
 
 A POA block header verifier should update its validator list according to the `voting_txs` field. With this design, we can verify headers and update the validator list without downloading the whole blocks.
+
+### Header verification
+
+We need to do some verification which slightly differs from the mainnet:
+
+`nonce` must set to all zero bytes.
+`compact_target` must set to `553648127`(difficulty 1) or `545259520`(difficulty 2) according to the attester index is whether equals to `N % VALIDATOR_COUNT` or not.
+
+The other fields keep using mainnet verification rules.
 
 ## references
 
--- a/rfcs/0000-poa-testnet/0000-poa-testnet.md
+++ b/rfcs/0000-poa-testnet/0000-poa-testnet.md
@@ -21,7 +21,7 @@ We expect the new POA testnet to be long term stable, and the testnet should be 
 
 In the purpose we design the POA testnet with the following principles:
 
-* There should be a dynamic group of validators to maintain the POA network, each validator can submit block and vote. With enough votes, the network can allow new validators to join or evict a exists validators.
+* There should be a dynamic group of validators to maintain the POA network, each validator can submit block and vote. With enough votes, the network can allow new validators to join or evict exists validators.
 * The network cannot be halted or concerned by minority malicious validators, malicious validators should be eventually evicted by majority honest validators.
 * The other part of the network should be the same as the mainnet(unless block header may need some extract context to do verification).
 
@@ -29,7 +29,8 @@ In the purpose we design the POA testnet with the following principles:
 
 We define the following variables:
 
-* `VALIDATOR_COUNT` - Number of current validators, this value changed due to new validators join or old validator leaves.
+* `MAX_VALIDATOR_COUNT` - Number of max validators.
+* `VALIDATOR_COUNT` - Number of current validators, this value changed due to new validators join or old validator leaves, but can't be more than `MAX_VALIDATOR_COUNT`.
 * `ATTEST_INTERVAL` - A validator cannot attest two blocks within `ATTEST_INTERVAL` number. For example, a validator who attest block (6) must wait for at least `ATTEST_INTERVAL` blocks to do next attest: block (6 + `ATTEST_INTERVAL` + 1). Notice when the `VALIDATOR_COUNT` <= `ATTEST_INTERVAL`, the POA testnet will stuck forever due to no validators can attest a new block.
 * `BLOCK_INTERVAL` - the interval of blocks, set to 8 seconds.
 * `BLOCK_TIMEOUT` - timeout for new block generation, set to 10 seconds.
@@ -54,11 +55,13 @@ Validators can use a simple strategy:
 
 Notice the difficulty set to `2` when an in-turn attester produces a block and set to `1` when a not in-turn attester produces a block. This makes the in-turn attester's block total difficulty higher than the not in-turn attester's block; once two validators attest at the same height due to the network error, the other nodes still chooses the higher total difficulty block as the main chain.
 
-`ATTEST_INTERVAL` is used to preventing malicious validators censor the POA network. A validator cannot attest two blocks within the `ATTEST_INTERVAL` number.  Consider the situation that malicious validators try to control the whole network: 
+`ATTEST_INTERVAL` is used to preventing malicious validators censor the POA network. A validator cannot attest two blocks within the `ATTEST_INTERVAL` number.  Consider the situation that malicious validators try to control the whole network:
 
 The first malicious validator must wait for `ATTEST_INTERVAL + 1` blocks to produce a block again. When the number of malicious validators is less than `ATTEST_INTERVAL + 1`, there must exist an honest validator who has the chance to submit a block includes majority voting to evict malicious validators.
 
 `ATTEST_INTERVAL` can be set to `VALIDATOR_COUNT / 2` the honest validators could eventually evict malicious validators unless the half of validators corrupted.
+
+One thing is that CKB uses 2-phase commitment, a transaction must be proposed first before committed in a block. This means the honest validators need at least produce two blocks to finally commit the eviction transaction, and these two blocks must be within the proposal window, so we choose a large enough value in the POA testnet: `TX_PROPOSAL_WINDOW` is set to `ProposalWindow(2, MAX_VALIDATOR_COUNT)`.
 
 ### validator list and on-chain governance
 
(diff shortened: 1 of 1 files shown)


rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,124 @@
...
+---
+
+# POA testnet
+
+## Motivation
+
+The original intention of aggron testnet is to provide a development-friendly testnet,
+ allow develops to deploy and test contracts, it's supposed to be similar to mainnet and provide stable services.
+
+However, the mining power on aggron is very unstable[chars](https://explorer.nervos.org/aggron/charts), Obviously for a reasonable miner there is no incitive to mine the testnet coins, and any new miner or mining pool join the testnet will speedup the block produces time temporally, then they leave, the testnet become extremely slow, the avg block time becomes few minutes or even longer.
```
stwith:
```suggestion
However, the mining power on aggron is very unstable (shown in [chars](https://explorer.nervos.org/aggron/charts)), obviously for a reasonable miner there is no incitive to mine the testnet coins, and any new miner or mining pool join the testnet will speedup the block produces time temporally, then they leave, the testnet become extremely slow, the avg block time becomes few minutes or even longer.
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,124 @@
...
+
+In the purpose we design the POA testnet with the following principles:
+
+* There should be a dynamic group of validators to maintain the POA network, each validator can submit block and vote. With enough votes, the network can allow new validators to join or evict a exists validators.
+* The network cannot be halted or concerned by minority malicious validators, malicious validators should be eventually evicted by majority honest validators.
+* The other part of the network should be the same as the mainnet(unless block header may need some extract context to do verification).
+
+## POA protocol
+
+For easy to describe the protocol, we define the following variables:
```
janx:
```suggestion
We define the following variables:
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,124 @@
...
+Validators use a round-robin style to attest blocks.
+For block `n`, the validator which index equals `n % VALIDATOR_COUNT` suppose to be the attester. However, a validator can fail to produce a block in time due to network congestion or other reasons, in this case, other validators should produce the block `n`.
+
+Validators can use a simple strategy to achieve this:
+
+* If `INDEX == n % VALIDATOR_COUNT`, wait for `BLOCK_INTERVAL` seconds then produce a new block with difficulty set to `2`.
+* If `INDEX != n % VALIDATOR_COUNT`, wait for `BLOCK_INTERVAL + rand(VALIDATOR_COUNT) * 0.5` seconds then produce a new block with difficulty set to `1`.
+* If the validator is the attester of the last block, wait for `ATTEST_INTERVAL` blocks then continue this strategy.
+
+`ATTEST_INTERVAL` is used for preventing malicious validators to censors the POA network. When malicious validators are less than `ATTEST_INTERVAL + 1` that at least an honest validator have the chance to submit votes and evict malicious validators.
```
janx:
```suggestion
`ATTEST_INTERVAL` is used for preventing malicious validators to censor the POA network. When malicious validators are less than `ATTEST_INTERVAL + 1` at least one honest validator has the chance to submit votes and evict malicious validators.
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,124 @@
...
+
+``` txt
+pubkey_hash_1 | pubkey_hash_2 | pubkey_hash_3 ...
+```
+
+We can use an `M of N` multisig lock to describe the validator list, the `M` is set to `VOTE_LIMIT`, and `N` is `VALIDATORS_COUNT`.
+
+To change the validator list, a validator can collect enough votes(the signatures) and send a tx to spent the old multisig lock and construct a new multisig lock with new pubkey hashes. A validator node can use `type_id` to track this validator list.
+
+The pre-defined [multisig script](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0021-ckb-address-format/0021-ckb-address-format.md#short-payload-format) is satisfied the requirements, one except is the `multisig script` should be public so other nodes can check new validator list. Validators should make sure the entire `multisig script` is stored to cell data, and only one cell is constructed to represent the validator list.
```
janx:
```suggestion
The pre-defined [multisig script](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0021-ckb-address-format/0021-ckb-address-format.md#short-payload-format) satisfies the requirements except that `multisig script` should be revealed so other nodes can check new validator list. Validators should make sure the entire `multisig script` is stored in a cell, and only one cell is constructed to represent the validator list.
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,124 @@
...
+
+Notice, the process of collect data is not included in the POA protocol for simplifying, a responsible validator should check the vote tx carefully before signing it.
```
janx:
```suggestion
Notice, the process of collecting votes is not included in the POA protocol for simplification, a responsible validator should check votes carefully before signing it.
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,124 @@
...
+
+## POA header
+
+Unfortunately, CKB header is fixed-length, the `nonce` field takes 128 bits, we can't just put 256 bits secp256k1 signature into the header.
```
janx:
```suggestion
CKB header is fixed-length, the `nonce` field is 128 bits, we can't put 256 bits secp256k1 signature into the header directly.
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,124 @@
...
+
+For easier implementation, instead change the block header structure, we put an extra POA context concatenated after cellbase transaction's first witness. To verify a header attestation, we need both the header and the cellbase transaction.
```
janx:
```suggestion
For the ease of implementation, instead of changing the block header structure, we put an extra POA payload `POAContext` in the cellbase transaction's first witness. To verify a header attestation, we need both the header and the cellbase transaction.
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,124 @@
...
+    merkle_proof:           Byte32Vec, // merkle proof of cellbase and voting txs
+    voting_txs:             TransactionVec,
+}
+```
+
+The first field `signature` is a secp256k1 signature signed by a validator.
+
+The signature message is supposed to digest the whole block except the signature itself.
+
+Because the signature is put into the cellbase transaction's witness, and the cellbase itself is digested by the `transaction_root` in the header. If we change the cellbase transaction the `transaction_root` and block hash will also be changed.
```
janx:
```suggestion
Because the signature is put into the cellbase transaction's witness, and the cellbase itself is digested by the `transaction_root` in the header. If we change the cellbase transaction the `transaction_root` and block hash will also be changed. A validator needs to 'zeroize' the signature field before signing.
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,183 @@
...
+Created: 2019-12-9
+---
+
+# POA testnet
+
+## Motivation
+
+The original intention of the aggron testnet is to provide a stable environment for contract development.
+
+However, the mining power on aggron is volatile (shown in [charts](https://explorer.nervos.org/aggron/charts)), since a rational miner has no incentive to mine the testnet coins continuously. If any new miner or mining pool joins, the mining difficulty rises fast. Once they leave, the testnet block produce becomes extremely slow due to a sudden drop of hashrate, and the average block time would require minutes or even longer.
```
janx:
```suggestion
However, the mining power on aggron is volatile (shown in [charts](https://explorer.nervos.org/aggron/charts)), since a rational miner has no incentive to mine the testnet coins continuously. If any new miner or mining pool joins, the mining difficulty rises fast. Once they leave, new block generation becomes extremely slow due to a sudden drop of hashrate, and the average block time would require minutes or even longer.
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,183 @@
...
+
+A POW blockchain only works when the miner has economic incentives, it's not adapted for the testnet, so we propose a new POA consensus testnet to serve the contract development purpose.
```
janx:
```suggestion
A POW blockchain only works when the miner has economic incentives that conflict with the nature of the testnet, so we propose a new POA consensus testnet to serve the contract development purpose.
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,183 @@
...
+* The other part of the network should be the same as the mainnet(unless block header may need some extract context to do verification).
+
+## POA protocol
+
+We define the following variables:
+
+* `VALIDATOR_COUNT` - Number of current validators, this value changed due to new validators join or old validator leaves.
+* `ATTEST_INTERVAL` - A validator cannot attest two blocks within `ATTEST_INTERVAL` number. For example, a validator who attest block (6) must wait for at least `ATTEST_INTERVAL` blocks to do next attest: block (6 + `ATTEST_INTERVAL` + 1). Notice when the `VALIDATOR_COUNT` <= `ATTEST_INTERVAL`, the POA testnet will stuck forever due to no validators can attest a new block.
+* `BLOCK_INTERVAL` - the interval of blocks, set to 8 seconds.
+* `VOTE_LIMIT` - The least votes to make a new validator join or to evict an old validator, should be at least `VALIDATOR_COUNT / 2 + 1`.
```
janx: `VOTE_THRESHOLD` or `VOTE_QUORUM` might be a better name.

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,183 @@
...
+
+The protocol uses `n % VALIDATOR_COUNT` as the index to choose the attester from validators; each validator checks the index and decides whether to attest a new block itself or wait for another attester to produce a new block.
+
+However, an in-turned attester may fail to produce a block due to network error or other reasons; in this case, other validators must produce the new block to make the POA testnet continue.
+
+Validators can use a simple strategy:
+
+1. For block height `n` a validator checks `n % VALIDAROR_COUNT`.
+2. If `INDEX == n % VALIDATOR_COUNT`, which means the validator is in it's turn to attests block `n`. The validator should wait for `BLOCK_INTERVAL` seconds then attests the block `n` with difficulty set to `2`.
+3. If `INDEX != n % VALIDATOR_COUNT`, which means the validator is not in it's turn to attests block `n`. The validator should wait for `BLOCK_INTERVAL + rand(VALIDATOR_COUNT) * 0.5` seconds to wait for another attester to produce a new block. If there are no new block produced during the time, the validator should attest a new block with difficulty set to `1`.
```
janx: Could the result of `BLOCK_INTERVAL + rand(VALIDATOR_COUNT) * 0.5` be `BLOCK_INTERVAL`? Or very close to `BLOCK_INTERVAL`? If so a validator and attester may generate a new block at nearly the same time, and the validator's block arrives other nodes earlier than the attester's. This could cause a lot of 1-block fork switch on testnet.

I suggest adding a buffer to validator's wait period, e.g. `BLOCK_INTERVAL * 2 + rand(VALIDATOR_COUNT) * 0.5`.

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,183 @@
...
+
+`ATTEST_INTERVAL` is used to preventing malicious validators censor the POA network. A validator cannot attest two blocks within the `ATTEST_INTERVAL` number.  Consider the situation that malicious validators try to control the whole network: 
+
+The first malicious validator must wait for `ATTEST_INTERVAL + 1` blocks to produce a block again. When the number of malicious validators is less than `ATTEST_INTERVAL + 1`, there must exist an honest validator who has the chance to submit a block includes majority voting to evict malicious validators.
+
+`ATTEST_INTERVAL` can be set to `VALIDATOR_COUNT / 2` the honest validators could eventually evict malicious validators unless the half of validators corrupted.
+
+### validator list and on-chain governance
+
+As previously described, the POA testnet maintained by a group of validators. We have a build-in on-chain governance mechanism to allows updating the exists validator list. The on-chain governance mechanism designed to be loose; it allows several validators that greater than `VOTE_LIMIT` to sign a transaction together that updating the exists validators to a list of new validators.
```
janx:
```suggestion
As previously described, the POA testnet maintained by a group of validators. We have a build-in on-chain governance mechanism to update the exists validator list. The on-chain governance mechanism designed to be loose; it allows several validators that greater than `VOTE_LIMIT` to sign a transaction together that updating the exists validators to a list of new validators.
```

janx: What does 'designed to be loose` mean here?

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,183 @@
...
+
+The pre-defined [multisig script] satisfies our on-chain governance requirements except that `multisig script` should be revealed so other nodes can check the new validator list.
+
+So we choose to use a cell to represent the POA validator list. The cell contains the following `multisig script` in its data field, and with a multisig lock to lock the cell, the `multisig script` in data must corresponding to the lock's `multisig script`.
+
+``` txt
+0 | 0 | VOTE_LIMIT | VALIDATOR_COUNT | blake160(Pubkey1) | blake160(Pubkey2) | ...
+```
+
+We use a cell with `M of N` multisig lock to describe the validator list, the`M` set to `VOTE_LIMIT`, and `N` set to `VALIDATORS_COUNT`, we need an extra parameter to represent `ATTEST_INTERVAL`, so the final data is a 4 bytes little-endian number to represent `ATTEST_INTERVAL` plus a `multisig script`:
```
janx:
```suggestion
We use a cell with `M of N` multisig lock to describe the validator list, the `M` set to `VOTE_LIMIT`, and `N` set to `VALIDATORS_COUNT`, we need an extra parameter to represent `ATTEST_INTERVAL`, so the final data is a 4 bytes little-endian number to represent `ATTEST_INTERVAL` plus a `multisig script`:
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,183 @@
...
+
+``` txt
+ATTEST_INTERVAL(4 bytes little-endian) | 0 | 0 | VOTE_LIMIT | VALIDATOR_COUNT | blake160(Pubkey1) | blake160(Pubkey2) | ...
+```
+
+To change the validator list, anyone that collects enough votes(the signatures) can send a transaction to update the old validator list cell and construct a new validator list cell.
+
+For simplify to track this cell, we assign a `type_id` on it; validators or anyone who wants to verify POA blocks must track this cell to keep validator list fresh.
```
janx:
```suggestion
To simplify the tracking of this cell, we assign a `type_id` on it; validators or anyone who wants to keep a fresh validator list can track the change of this cell by its `type_id`.
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,183 @@
...
+
+Before sign a vote transaction, validators must make sure the `multisig script`  in the cell is valid and corresponding to the lock, and only one cell constructed to represent the validator list; otherwise, the POA consensus wound broken.
```
janx:
```suggestion
Before signing a vote transaction, validators must make sure the `multisig script` in the cell is valid and corresponding to the lock, and only one cell constructed to represent the validator list; otherwise, the POA consensus wound broken.
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,183 @@
...
+
+To simplify, the process of collecting votes do not include in the POA protocol.
```
janx:
```suggestion
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,183 @@
...
+
+### off-chain governance
+
+The main intention of this document is to describe the POA testnet protocol. However, to better explain how the POA testnet validator works, we add this section as a suggestion to the off-chain governance, notice this section is more like a suggestion than a specification.
+
+The intention of on-chain governance and dynamic validators design is to decentralize the POA testnet. Instead of the foundation itself, the community should finally maintain the POA testnet validators.
```
janx:
```suggestion
The intention of on-chain governance and dynamic validators design is to decentralize the POA testnet. Instead of the foundation, the community should be the maintainer of the POA testnet.
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,183 @@
...
+
+In the initial stage of the POA testnet, there may be just a small group validator that invited by the Nervos foundation. The validators together should governance the testnet and invite more validators from the community. A candidate validator may be a company or organization that influences the community.
```
janx:
```suggestion
In the initial stage of the POA testnet, there may be just a small group validator that invited by the Nervos foundation. The validators together should governance the testnet and invite more validators from the community. A candidate validator may be a reputable company or organization in the community. 
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,183 @@
...
+
+The main intention of this document is to describe the POA testnet protocol. However, to better explain how the POA testnet validator works, we add this section as a suggestion to the off-chain governance, notice this section is more like a suggestion than a specification.
+
+The intention of on-chain governance and dynamic validators design is to decentralize the POA testnet. Instead of the foundation itself, the community should finally maintain the POA testnet validators.
+
+In the initial stage of the POA testnet, there may be just a small group validator that invited by the Nervos foundation. The validators together should governance the testnet and invite more validators from the community. A candidate validator may be a company or organization that influences the community. 
+
+A possible way to elect candidate validators: 
+
+The exists validators deploy a permissionless voting DAPP on the mainnet allows any company and organization can register them as a candidate.
```
janx:
```suggestion
The exists validators deploy a permissionless voting DAPP on the mainnet allows any company and organization to register as a candidate. 
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,183 @@
...
+
+The exists validators deploy a permissionless voting DAPP on the mainnet allows any company and organization can register them as a candidate. 
+Anyone can deposit mainnet coins to vote on the candidates. 
+The POA testnet validators obey the result of the DApp to invites the candidate to become the testnet validator.
+
+Such an election should periodically host.
+
+## POA header
+
+CKB header encoded in fixed-length, the size of the `nonce` field is 128 bits, we can't put 256 bits secp256k1 signature into the header directly.
```
janx:
```suggestion
CKB header is length fixed, the size of the `nonce` field is 128 bits, we can't put 256 bits secp256k1 signature into the header directly.
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,183 @@
...
+    2. calculate `transactions_root = merkle_root([raw_transactions_root, witness_transactions_root])`
+    3. check `transactions_root` equals to `header.transactions_root`
+2. zero `signature` field of POAContext:
+    1. extract `POAContext` from cellbase transaction
+    2. set `signature` field to "0x000...0000"
+    3. set `POAContext` back to cellbase transaction witness
+3. re-comute the block's `transaction_root` and block hash. 
+4. recovery the pubkey, then check the pubkey hash is belongs to a validator.
+
+Now we can verify a POAHeader with fixed validators. However, when validator list updated, the header-first synchronization may couldn't handle it correctly, think the following situation:
```
janx:
```suggestion
Now we can verify a POAHeader with fixed validators. However, when validator list updated, the header-first synchronization may couldn't handle it correctly, consider the following situation:
```

rfcs/0000-poa-testnet/0000-poa-testnet.md
jjyr: It means that the on-chain governance contract has very few verification rules, I think I can just remove this.

rfcs/0000-poa-testnet/0000-poa-testnet.md
jjyr: We could add another variable `BLOCK_TIMEOUT`, set a value that slightly greater than `BLOCK_INTERVAL` like `10s`.

Every validator who not in its turn waits for `BLOCK_TIMEOUT + rand(VALIDATOR_COUNT) * 0.5` seconds to produce a block, this reduces the possibilities that two validators attest in the same time.

In fact, even there are two blocks produced at the same time it's won't hurt; the block that attested by an in-turn validator always has higher `total difficulty`; even if two blocks have same `total difficulty`, the next in turned validator can decide the main chain, the chain selection is just like the POW.

nirenzang:
I am curious when validator "n % VALIDATOR_COUNT" did not produce a block, why can't we ask the next validator to produce two blocks in a row? That would be much simpler, and we have no risk of short forks.

rev-chaos:
I'm not sure this mechanism could ensure converge in high probability. Forks happen easily,  a validator attests for different forks in ATTEST_INTERVAL could not be considered malicious, or the consensus may halt. Then malicious validators can make use of this to break converge.

rev-chaos:
A suggestion that might improve convergence is to set the weight of a block which attests by attester to be 2 times the previous block's.

rev-chaos:
A suggestion that might improve convergence is to set the weight of a block which attests by an attester to be 1 and the one attests by a non-attester to be 0.

shuyang-sjtu:
I am also curious about the case when the corresponding validator of a round fails to propose a block. It is hard to catch the intuition behind the strategy. For example, what role the randomness plays in "rand(VALIDATOR_COUNT) * 0.5" and why the difficulty is set to "2 when an in-turn attester produces a block and set to 1 when a not in-turn attester produces a block". Is it possible that validators not-in-turn collude to propose a chain of continuous blocks of weight 1? I know this is what ATTEST_INTERVAL and the eviction rule are set for. However, for the eviction rule, it is hard to formally (by a program) tell adversaries from nodes suffering delays. I am worried that there can still be an attack by continuous 1-weight blocks despite the ATTEST_INTERVAL since adversary nodes may "pipeline" and the ATTEST_INTERVAL could not be too long since VALIDATOR_COUNT > ATTEST_INTERVAL must hold.

janx:
It seems some questions are asked due to inconsistent views on the security model of testnet PoA:

- what assumptions can we make?
- what kind of adversarial behaviors should be tolerated?
- in what circumstances testnet should be reset?

jjyr:
> I am also curious about the case when the corresponding validator of a round fails to propose a block. It is hard to catch the intuition behind the strategy. For example, what role the randomness plays in "rand(VALIDATOR_COUNT) * 0.5" and why the difficulty is set to "2 when an in-turn attester produces a block and set to 1 when a not in-turn attester produces a block". Is it possible that validators not-in-turn collude to propose a chain of continuous blocks of weight 1? I know this is what ATTEST_INTERVAL and the eviction rule are set for. However, for the eviction rule, it is hard to formally (by a program) tell adversaries from nodes suffering delays. I am worried that there can still be an attack by continuous 1-weight blocks despite the ATTEST_INTERVAL since adversary nodes may "pipeline" and the ATTEST_INTERVAL could not be too long since VALIDATOR_COUNT > ATTEST_INTERVAL must hold.

1. The randomness "rand(VALIDATOR_COUNT) * 0.5" is for reducing the possibility that two validators produce a block at the same time when an attester skips its turn.
2. The reason of set difficulty to `2` for an in-turn attestation, is that due to the desynchronize clock and network delay, a validator may not see the new block then attests a not in-turn block; in this case, the in-turn block has higher total-difficulty, so client more likely choose the in-turn block as the main chain.
3. This protocol assumes the number of adversary nodes less than half of `VALIDATOR_COUNT`, so we can set `ATTEST_INTERVAL` to `VALIDATOR_COUNT / 2`,  so the eviction couldn't be stopped under this assumption.

jjyr:
> I'm not sure this mechanism could ensure converge in high probability. Forks happen easily, a validator attests for different forks in ATTEST_INTERVAL could not be considered malicious, or the consensus may halt. Then malicious validators can make use of this to break converge.

The intention of this protocol is not to force converge to a single chain; instead of eliminating the fork, we eliminate the validators who make the fork; the core part of this protocol is to make sure that honest validators eventually can perform an eviction to remove malicious validators from the validator list.

shuyang-sjtu:
> > I am also curious about the case when the corresponding validator of a round fails to propose a block. It is hard to catch the intuition behind the strategy. For example, what role the randomness plays in "rand(VALIDATOR_COUNT) * 0.5" and why the difficulty is set to "2 when an in-turn attester produces a block and set to 1 when a not in-turn attester produces a block". Is it possible that validators not-in-turn collude to propose a chain of continuous blocks of weight 1? I know this is what ATTEST_INTERVAL and the eviction rule are set for. However, for the eviction rule, it is hard to formally (by a program) tell adversaries from nodes suffering delays. I am worried that there can still be an attack by continuous 1-weight blocks despite the ATTEST_INTERVAL since adversary nodes may "pipeline" and the ATTEST_INTERVAL could not be too long since VALIDATOR_COUNT > ATTEST_INTERVAL must hold.
> 
> 1. The randomness "rand(VALIDATOR_COUNT) * 0.5" is for reducing the possibility that two validators produce a block at the same time when an attester skips its turn.
> 2. The reason of set difficulty to `2` for an in-turn attestation, is that due to the desynchronize clock and network delay, a validator may not see the new block then attests a not in-turn block; in this case, the in-turn block has higher total-difficulty, so client more likely choose the in-turn block as the main chain.
> 3. This protocol assumes the number of adversary nodes less than half of `VALIDATOR_COUNT`, so we can set `ATTEST_INTERVAL` to `VALIDATOR_COUNT / 2`,  so the eviction couldn't be stopped under this assumption.

For 1, I am totally convinced. For 2-3, what I meant is I am concerned that "1" weight for not-in-turn blocks might be too great (it might look more natural if it is something parameterized by ATTEST_INTERVAL / VALIDATOR_COUNT), here is an example (which tells what I meant by "pipeline") that the adversary could forever propose 1-weight blocks with half of all nodes being adversaries.
           Round1   Round2  Round3  Round4  Round5  Round6  Round7
adv1       1                                                1                                               1
adv2                        1                                              1
adv3                                       1                                               1
Here we assume VALIDATOR_COUNT=6, ATTEST_INTERVAL=3 and 3 of them are malicious  (sure the assumption is "<1/2 VALIDATOR_COUNT" rather than "<=1/2" so it might be somehow different). "1" (in the table) for Round_i and adv_j means the j-th adversary proposes a 1-weight block in round i. In this case, malicious nodes can cooperate to propose a chain of continuous 1-weight blocks of any large total weight they want. But I am not sure whether this will be a great problem to worry about, since this issue can be possibly solved by the eviction.

shuyang-sjtu:
> > I am also curious about the case when the corresponding validator of a round fails to propose a block. It is hard to catch the intuition behind the strategy. For example, what role the randomness plays in "rand(VALIDATOR_COUNT) * 0.5" and why the difficulty is set to "2 when an in-turn attester produces a block and set to 1 when a not in-turn attester produces a block". Is it possible that validators not-in-turn collude to propose a chain of continuous blocks of weight 1? I know this is what ATTEST_INTERVAL and the eviction rule are set for. However, for the eviction rule, it is hard to formally (by a program) tell adversaries from nodes suffering delays. I am worried that there can still be an attack by continuous 1-weight blocks despite the ATTEST_INTERVAL since adversary nodes may "pipeline" and the ATTEST_INTERVAL could not be too long since VALIDATOR_COUNT > ATTEST_INTERVAL must hold.
> 
> 1. The randomness "rand(VALIDATOR_COUNT) * 0.5" is for reducing the possibility that two validators produce a block at the same time when an attester skips its turn.
> 2. The reason of set difficulty to `2` for an in-turn attestation, is that due to the desynchronize clock and network delay, a validator may not see the new block then attests a not in-turn block; in this case, the in-turn block has higher total-difficulty, so client more likely choose the in-turn block as the main chain.
> 3. This protocol assumes the number of adversary nodes less than half of `VALIDATOR_COUNT`, so we can set `ATTEST_INTERVAL` to `VALIDATOR_COUNT / 2`,  so the eviction couldn't be stopped under this assumption.

For 1, I am totally convinced. For 2-3, what I meant is I am concerned that "1" weight for not-in-turn blocks might be too great (it might look more natural if it is something parameterized by ATTEST_INTERVAL / VALIDATOR_COUNT), here is an example (which tells what I meant by "pipeline") that the adversary could forever propose 1-weight blocks with half of all nodes being adversaries.

  | Round1 | Round2 | Round3 | Round4 | Round5 | Round6 | Round7 |  
Adv1 | 1 | | | 1 | | | 1 |
Adv2 | | 1 | | | 1 | | |
Adv3 | | | 1 | | | 1 | |

Here we assume VALIDATOR_COUNT=6, ATTEST_INTERVAL=3 and 3 of them are malicious (sure the assumption is "<1/2 VALIDATOR_COUNT" rather than "<=1/2" so it might be somehow different). "1" (in the table) for Round_i and adv_j means the j-th adversary proposes a 1-weight block in round i. In this case, malicious nodes can cooperate to propose a chain of continuous 1-weight blocks of any large total weight they want. But I am not sure whether this will be a great problem to worry about, since this issue can be possibly solved by the eviction.

jjyr:
> It seems some questions are asked due to inconsistent views on the security model of testnet PoA:
> 
> * what assumptions can we make?
> * what kind of adversarial behaviors should be tolerated?
> * in what circumstances testnet should be reset?

* what assumptions can we make?
  1. The number of adversary nodes less than half of `VALIDATOR_COUNT`
* what kind of adversarial behaviors should be tolerated?
  1. Since our purpose is to support testnet: censorship, collude, mined on forks won't hurt real assets, all these behaviors won't halt the chain either and made the malicious nodes easy to be detected by a program.
* in what circumstances testnet should be reset?
  1. malicious nodes are more than half of `VALIDATOR_COUNT`
  2. validators can't make an agreement of eviction on off-chain governance.

jjyr:
@shuyang-sjtu 

https://github.com/nervosnetwork/rfcs/pull/158/files#diff-6eca90ec8afbdba69e3e0b53a6dcae4dR33

After produces block 1, Adv1 must wait for `ATTEST_INTERVAL` to produce the next block which number is 5.

So Adv2 and Adv3 can't produce block `2, 3, 4` continuously; there must be at least one block within `2, 3, 4` that produced by an honest validator; honest validators should use this chance to evict adversaries.

shuyang-sjtu:
> @shuyang-sjtu
> 
> https://github.com/nervosnetwork/rfcs/pull/158/files#diff-6eca90ec8afbdba69e3e0b53a6dcae4dR33
> 
> After produces block 1, Adv1 must wait for `ATTEST_INTERVAL` to produce the next block which number is 5.
> 
> So Adv2 and Adv3 can't produce block `2, 3, 4` continuously; there must be at least one block within `2, 3, 4` that produced by an honest validator; honest validators should use this chance to evict adversaries.

I know if "ATTEST_INTERVAL <1/2 VALIDATOR_COUNT", the adversary cannot propose block 4, and that is what I meant by "_sure the assumption is "<1/2 VALIDATOR_COUNT" rather than "<=1/2" so it might be somehow different_" if you read by assuming I am not stupid.

jjyr:
> I am curious when validator "n % VALIDATOR_COUNT" did not produce a block, why can't we ask the next validator to produce two blocks in a row? That would be much simpler, and we have no risk of short forks.

If we allow the validator to produce n blocks, how to distinguish that if there really n blocks are skipped or the validator is malicious? for example:

For a 4 validator case, A, B is malicious, and C, D is honest. after B produces a block, A can pretend that C and D do not produce blocks within timeout, how C and D(or other nodes) handling this situation?

nirenzang:
> > I am curious when validator "n % VALIDATOR_COUNT" did not produce a block, why can't we ask the next validator to produce two blocks in a row? That would be much simpler, and we have no risk of short forks.
> 
> If we allow the validator to produce n blocks, how to distinguish that if there really n blocks are skipped or the validator is malicious? for example:
> 
> For a 4 validator case, A, B is malicious, and C, D is honest. after B produces a block, A can pretend that C and D do not produce blocks within timeout, how C and D(or other nodes) handling this situation?

C and D will keep producing blocks at their designated time slot, and their blocks will have heavier weights, so they will consider the chain constructed by themselves the authentic chain. Am I missing something? I am not familiar with POA so very likely I miss sth.

nirenzang:
I see. As long as we 
1. log all info on the testnet consensus (block proposer, block receiving time, vote on eliminating block proposer...), 
2. raise an alarm when sth goes wrong (chain growth rate significantly lower than expected, number of orphaned blocks significantly higher than expected, blocks proposed by some proposers significantly larger than the others),

we can always solve the problem off-line when an alarm is raised via manual examination.

shuyang-sjtu:
This protocol is extremely suitable for testing smart contracts on the testnet. As a matter of fact, it is very unlikely to have a great portion of adversaries in the real testnet since they have no incentive to undermine the system for development. This protocol can rule out all attacks come up by me even with almost half total validator nodes malicious, as long as an alarm is raised in time.

rfcs/0000-poa-testnet/0000-poa-testnet.md
```diff
@@ -0,0 +1,186 @@
...
+
+`ATTEST_INTERVAL` is used to preventing malicious validators censor the POA network. A validator cannot attest two blocks within the `ATTEST_INTERVAL` number.  Consider the situation that malicious validators try to control the whole network:
+
+The first malicious validator must wait for `ATTEST_INTERVAL + 1` blocks to produce a block again. When the number of malicious validators is less than `ATTEST_INTERVAL + 1`, there must exist an honest validator who has the chance to submit a block includes majority voting to evict malicious validators.
+
+`ATTEST_INTERVAL` can be set to `VALIDATOR_COUNT / 2` the honest validators could eventually evict malicious validators unless the half of validators corrupted.
+
+One thing is that CKB uses 2-phase commitment, a transaction must be proposed first before committed in a block. This means the honest validators need at least produce two blocks to finally commit the eviction transaction, and these two blocks must be within the proposal window, so we choose a large enough value in the POA testnet: `TX_PROPOSAL_WINDOW` is set to `ProposalWindow(2, MAX_VALIDATOR_COUNT)`.
```
doitian: Nice catch

doitian:
Closed. See #194

PR #185, opened 2020-04-26
Add new code_hash_index to short address table

CipherWang:
add anyone-can-pay lock convention.

<p align="center">
  <a href="https://github.com/nervosnetwork/rfcs/compare/master...CipherWang:master?short_path=c520475#diff-c520475c6cf58d00caeb60d5883c5f31">[PREVIEW]</a>
</p>


quake:
I think this change should be held until any-one-can-pay rfc is approved

CipherWang:
> I think this change should be held until any-one-can-pay rfc is approved

Right, just list it here for the convenience of some projects that may use it.

PR #199, opened 2020-07-27
Support type script in CKB address

katat:
This PR is a proposal to have current CKB address RFC support the encoding of type script in addition to lock script. 

## Reasoning
In some use cases, such as creating an asset account on Neuron, where the metadata of a type script is partially hardcoded while the users need to pass in the `args` to identify an sUDT.

The example below is the hardcoded part in Neuron to handle sUDT transactions or related queries via Lumos indexer(ckb-indexer):

```
SUDT_SCRIPT_CODEHASH=0x48dbf59b4c7ee1547238021b4869bceedf4eea6b43772e5d66ef8865b6ae7212
SUDT_SCRIPT_HASHTYPE=data
SUDT_DEP_TXHASH=0xc1b2ae129fad7465aaa9acc9785f842ba3e6e8b8051d899defa89f5508a77958
SUDT_DEP_INDEX=0
SUDT_DEP_TYPE=code
```

When making queries via Lumos indexer, it only needs to specify the type script (`CODEHASH`, `HASHTYPE`, `ARGS`). Once the CKB address support encoding of type script, the applications won't need to hardcode the (`CODEHASH`, `HASHTYPE`) anymore, as the users can specify in application-level which type script they want to query with via a CKB address.

## Further thoughts
However, in order to transact with the sUDT, it will need the other metadata (`DEP_TXHASH`, `DEP_INDEX`, `DEP_TYPE`) to locate the **seed cell** holding the code represented by a type script. This applies to the case of anyone-can-pay lock script as well.

It is straightforward to update the existing CKB address RFC to support the encoding of type script, since the encoding rule is almost the same as lock script. For the `cell dependency` part, it might be worthwhile to consider if CKB address should be upgraded to support the encoding of `cell dependency` metadata. 

The benefit of doing so is that the applications won't need to hardcode any metadata for the users to transact with any UDT that can be specified with a single CKB address. Additionally, a CKB address can be referenced to locate a seed cell, which we don't seem to have a convenient way to locate at the moment. 

The downside is that it would make the CKB address string longer and handle the additional data points as it comes with the additional `cell dependency` information.

Therefore, welcome any comments and see if it makes sense to locate the **seed cell** using a CKB address.

### DIFF
--- a/rfcs/0021-ckb-address-format/0021-ckb-address-format.md
+++ b/rfcs/0021-ckb-address-format/0021-ckb-address-format.md
@@ -11,11 +11,14 @@ Created: 2019-01-20
 
 ## Abstract
 
-*CKB Address Format* is an application level cell **lock script** display recommendation. The lock script consists of three key parameters, including *code_hash*, *hash_type* and *args*. CKB address packages lock script into a single line, verifiable, and human read friendly format.
+*CKB Address Format* is recommended to handle the encodings for both **lock script** and **type script** in application level. CKB address can package a script into a single line format, which is verifiable and human-readable.
 
 ## Data Structure
 
-### Payload Format Types
+Both **lock script** and **type script** consist of three key properties, including *code_hash*, *hash_type* and *args*. Following the payload formatting rules outlined below, a script data structure can be encoded as a CKB address, which can be parsed for its original script reversely.
+
+### Lock script
+#### Payload Format Types
 
 To generate a CKB address, we firstly encode lock script to bytes array, name *payload*. And secondly, we wrap the payload into final address format.
 
@@ -27,7 +30,7 @@ There are several methods to convert lock script into payload bytes array. We us
 |  0x02       | full version with hash_type = "Data"           |
 |  0x04       | full version with hash_type = "Type"           |
 
-### Short Payload Format
+#### Short Payload Format
 
 Short payload format is a compact format which identifies common used code_hash by 1 byte code_hash_index instead of 32 bytes code_hash.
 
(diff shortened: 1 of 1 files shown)


katat:
One way, I can think of at the moment, to locate the **seed cell** without encoding the `cell dependency` in the address is as below:

1. Query live cells using lumos indexer with a type script.
2. Derive a corresponding dependency cell of the type script from any one of the live cells returned from the query.

This workaround would avoid the need to hardcode the metadata of a **seed cell** in the transaction generations, increasing the flexibility to support various types of contracts in the application level.

On the other hand, we will need to review the potential security issues in this workaround. One potential issue, by following this path, is the `type script` is not guaranteed to map to a desired **dependency cell** although it stores the same code. For example:

1. Alice deploys a type script T, which has a dependency cell D1
2. Bob deploys the same type script T, which has a dependency cell D2
3. Charlie makes a transaction involving T contract, referencing the D1 as the dependency cell.
4. Alice destroys D1

In this scenario, if Charlie wants to make another transaction involving T, the transaction would reference the D2 as the dependency cell. I am wondering if this would introduce any potential security risks.

doitian:
I don't think CKB address should encode type script. For example, if I have a mailing address, I can use it to receive various kinds of packages, I don't need an address to receive letters, and another to receive boxes.

katat:
Thanks @doitian for the comment.

Let me bring in a bit more details from where this proposal originated. The goal is mainly about achieving a better user experience around the uses of UDT or other protocols in the future. 

So far we got the sUDT integrated on Neuron, in which the users must fill in the `type_script.args` while the application hardcode the other fields of `type_script` and its corresponding fields in `cell_dep`. There are two problems with this approach:

1. It is not flexible enough for the applications to support other UDTs in the future since the UDT related data points are hardcoded. Meanwhile, it would imply _centralization_ if the application won't be able to allow the users to use any other UDTs they want in the blockchain world. 

2. The `type_script.args` is too technical for normal users to understand. The meaning of this field is very likely to be different in different type scripts, easy to get users confused.

It would be much more convenient to have a standard way to look up the relevant data points in a type script, so as to support consistent user experiences across different applications, such as [querying with the hash of a type script on explorer](https://explorer.nervos.org/aggron/sudt/0x1c5f32c5efb08ac256bd9413c30eef7420ae54f30ff11075967b89570326c295). 

In the current implementation of Explorer, it is based on the hash of a type script. We were trying to resolve this issue using the hash of type script with the additional support of lumos indexer to locate the type script metadata, so that users can pass around that same string, which the application can reference to look up the necessary data points for further queries through indexer or generating transactions involving the **contract script**.

However, after some conversations in the [thread](https://github.com/nervosnetwork/lumos/issues/19), it seems to make more sense to have CKB address support encoding for type script. 

From my view (if I got it right), the CKB address plays the role in packaging **scripts** into strings easily sharable. It currently supports the **lock script**. Thus, it seems cohesive to support **type script** as well.

katat:
I close this PR, as there is [a better approach](https://github.com/nervosnetwork/rfcs/pull/200) to resolve this script usability issue.

PR #200, opened 2020-07-29
[DRAFT] Cell mapper

katat:
This proposal is an alternative solution to the script usability issue that [my previous PR](https://github.com/nervosnetwork/rfcs/pull/199) was trying to resolve, as a result of the discussions with the core team members. 

The idea is to come up with a standard to reference an outpoint to locate a cell in a transaction, which has the native references to locate all the necessary metadata of a script. Thus, the new approach should be able to resolve the outstanding issues mentioned in my previous PR. With the ability to load script metadata on the fly, the applications would have more flexibility in the support of custom contract scripts in the long run.

For now, I name this RFC as **Cell Mapper**, because it would serve as a map from a cell to the metadata of a script. The alternative name could be **Script Mapper**. I am open-minded to any other suggestions in the naming.

As outlined below, the mapping process will involve two steps:
1. Encodings between an outpoint data structure and a cell address string.
2. A rule to fetch the script metadata relating to a cell.

It will include a CLI demo for everyone to play around the decode/encode of cell address, and query on the script metadata.

Please feel free to let me know if any comments.

### DIFF
--- a/rfcs/0025-cell-mapper/0025-cell-mapper.md
+++ b/rfcs/0025-cell-mapper/0025-cell-mapper.md
@@ -1,0 +1,26 @@
+---
+Number: "0025"
+Category: Standards Track
+Status: Proposal
+Author: Katat Choi
+Organization: Nervos Foundation
+Created: 2020-7-30
+---
+
+# CKB Cell Mapper
+
+## Abstract
+
+## Motivation
+
+## Specification
+
+### Cell Address Encoding
+
+### Collect Script Metadata 
+#### UDT type script
+#### ACP lock script
+
+## Example
+### Sample code
+### Demo repository


katat:
Hi @quake, 

Here is just draft. I think it would helpful for you and others to take a look at the overview before I work on the details outlined in the draft.

By the way, I can't change the PR reviewers. Please feel free to add in the other team members as reviewers. I guess @CipherWang @doitian @xxuejie would be interested.

Thanks,
Kata

quake:
May I know the general idea of `Cell Address Encoding`, does it mean encoding the outpoint (tx_hash | index) to a human readable address, like base58 encoding?

katat:
Yes, I think that is a critical question to answer before working on this RFC.

I am thinking to follow the bech32 algorithm of CKB address to do the encoding for the cell addresses. Also, I like the idea of having a prefix, `ckb` or `ckt`, to indicate whether an address targets the mainnet or testnet.

If this approach of encoding for cell addresses makes sense, then I am wondering if it would be better to extend the current CKB address RFC to support the cell address by having an additional format type, such as `0x10` in addition to its current supported types for lock script. So the CKB address standard could be referenced for all the different kinds of `addresses`. Otherwise, it could be confusing when having different RFCs for different addresses.

katat:
Close this for now. Will reopen this if there are more obvious use cases demanding for this feature on CKB.

PR #207, opened 2020-09-29
Update 0021-ckb-address-format.md

Keith-CY:
Fix the hash type of anyone_can_pay script

### DIFF
--- a/rfcs/0021-ckb-address-format/0021-ckb-address-format.md
+++ b/rfcs/0021-ckb-address-format/0021-ckb-address-format.md
@@ -37,11 +37,11 @@ payload = 0x01 | code_hash_index | args
 
 To translate payload to lock script, one can convert code_hash_index to code_hash and hash_type with the following *popular code_hash table*. And args as the args.
 
-| code_hash_index |        code_hash     |   hash_type  |          args           |
-|:---------------:|----------------------|:------------:|-------------------------|
-|      0x00       | SECP256K1 + blake160 |     Type     |  blake160(PK)*          |
-|      0x01       | SECP256K1 + multisig |     Type     |  multisig script hash** |
-|      0x02       | anyone_can_pay       |     Type     |  blake160(PK)           |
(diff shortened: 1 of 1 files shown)


rfcs/0021-ckb-address-format/0021-ckb-address-format.md
```diff
@@ -37,11 +37,11 @@ payload = 0x01 | code_hash_index | args
...
-| code_hash_index |        code_hash     |   hash_type  |          args           |
-|:---------------:|----------------------|:------------:|-------------------------|
-|      0x00       | SECP256K1 + blake160 |     Type     |  blake160(PK)*          |
-|      0x01       | SECP256K1 + multisig |     Type     |  multisig script hash** |
-|      0x02       | anyone_can_pay       |     Type     |  blake160(PK)           |
+| code_hash_index |        code_hash     |     hash_type        |          args           |
+|:---------------:|----------------------|:--------------------:|-------------------------|
+|      0x00       | SECP256K1 + blake160 |     Type             |  blake160(PK)*          |
+|      0x01       | SECP256K1 + multisig |     Type             |  multisig script hash** |
+|      0x02       | anyone_can_pay       |     Data/Type***     |  blake160(PK)           |
```
quake, line 44: Because the address rfc definition is split into two steps: 
1. script => payload
2. payload => address

```
address = hrp | separator | script_to_payload(script) | checksum
```

The first step (fn script_to_payload) has nothing to do with hrp as defined by rfc, and the generated payload is the same whether the script is used in mainnet or otherwise, and this change breaks that rule.

I vote -1 for this PR

quake (changes requested):

rfcs/0021-ckb-address-format/0021-ckb-address-format.md
doitian: So, what's your suggestion?

Keith-CY:
Now they are both of `type` hash type

PR #215, opened 2020-12-15
chore: fix typo and add link

shaojunda:

PR #217, opened 2021-01-18
load header immature rule

doitian:
- Add load header immature rule
- Syscall `load_header` can also load via dep cell.

The second commit is unrelated. I found it when adding the immature rule. They are both minor changes so I submit them together in one PR.

### DIFF
--- a/rfcs/0009-vm-syscalls/0009-vm-syscalls.md
+++ b/rfcs/0009-vm-syscalls/0009-vm-syscalls.md
@@ -445,7 +445,7 @@ In case of errors, `addr` and `index` will not contain meaningful data to use.
 #### Loading Header Immature Rule
 [loading header immature Rule]: #loading-header-immature-error
 
-**Attention** that the script can only load the header of a block which is 4 epochs ago, otherwise the transaction is immature and must wait. For example, if the block is the first block in epoch 4, a transaction loading its header can only be included in the first block of epoch 8 and later blocks.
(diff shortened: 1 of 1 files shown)


xxuejie (dismissed):

rfcs/0009-vm-syscalls/0009-vm-syscalls.md
```diff
@@ -441,6 +442,11 @@ This syscall might return the following errors:
 
 In case of errors, `addr` and `index` will not contain meaningful data to use.
 
+#### Loading Header Immature Rule
+[loading header immature Rule]: #loading-header-immature-error
+
+**Attention** that the script can only load the header of a block which is 4 epochs ago, otherwise the transaction is immature and must wait. For example, if the block is the first block in epoch 4, a transaction loading its header can only be included in the first block of epoch 8 and later blocks.
```
janx, line 448:
```suggestion
**Attention** that the script can only load the header of a block which is 4 epochs ago, otherwise the header is immature and the transaction must wait. For example, if the block is the first block in epoch 4, a transaction loading its header can only be included in the first block of epoch 8 and later blocks.
```

janx (changes requested):

janx (approved):

quake (dismissed):

doitian:
Resolved conflicts with https://github.com/nervosnetwork/rfcs/pull/215/files#diff-587e8c89cbf08dcef8f110b72409bd48e2548ddd58eae0754e724c288ead5cb6R218

janx (approved):

zhangsoledad (approved):

PR #222, opened 2021-02-03
RFC29: Allow multiple matches on dep cells via type script hash when these cells have the same data

doitian:


rfcs/0000-allow-script-multiple-matches-on-identical-code/0000-allow-script-multiple-matches-on-identical-code.md
```diff
@@ -0,0 +1,96 @@
...
+## Abstract
+
+This document proposes a transaction verification consensus change to allow multiple cell dep matches on type script hash when all the matches are resolved to the same script code.
+
+## Motivation
+
+Currently, CKB does not allow multiple matches when resolving the script by type script hash in dep cells. There are two scenarios in which this can cause troubles:
+
+1. Dep group is a feature to let transactions add multiple dep cells together. When a user adds multiple dep groups, these groups may have duplicated cells. The user has to expand the dep groups and remove the duplicates, which leads to a larger transaction.
+2. In protocols like Open Transaction, the transaction is assembled by merging several partial transactions. If these partial transactions contain duplicated dep groups, it will make the signature algorithm very complex.
```
keroro520, line 21: Currently, [CKB doesn't allow duplicated dep groups](https://github.com/nervosnetwork/ckb/blob/bea2fa3e0fa7cbf33a0f27f31cdaf9df348995ac/verification/src/transaction_verifier.rs#L405-L413). Do you mean "overlapping dep groups"?

doitian, line 21:
```suggestion
2. In protocols like Open Transaction, the transaction is assembled by merging several partial transactions. If these partial transactions contain overlapping dep groups, it will make the signature algorithm very complex.
```

yangby-cryptape (dismissed):

rfcs/0222-allow-script-multiple-matches-on-identical-code/0222-allow-script-multiple-matches-on-identical-code.md
```diff
@@ -1 +1,93 @@
-In review, see <https://github.com/nervosnetwork/rfcs/pull/222>
+---
+Number: "0222"
```
janx, line 2: There's never been any formal discussion on changing numbering rules, I think we should stick to the existing rule.

janx (changes requested):

yangby-cryptape (dismissed):

rfcs/0029-allow-script-multiple-matches-on-identical-code/0029-allow-script-multiple-matches-on-identical-code.md
```diff
@@ -1 +1,93 @@
...
+Created: 2021-02-03
+---
+
+# Allow Multiple Cell Dep Matches When There Is No Ambiguity
+
+## Abstract
+
+This document proposes a transaction verification consensus change to allow multiple cell dep matches on type script hash when all the matches are resolved to the same script code.
+
+## Motivation
```
xxuejie, line 16: I would recommend a re-org in this section:

* First we talked about the fact that there are 2 ways to reference a cell
* Now we introduce the problem: multiple matches for a single type script hash would be an error in current CKB(possibly an example to illustrate this), then we can argue the legitimacy of allowing multiple matches

rfcs/0029-allow-script-multiple-matches-on-identical-code/0029-allow-script-multiple-matches-on-identical-code.md
```diff
@@ -1 +1,93 @@
...
+
+Currently, CKB does not allow multiple matches when resolving the script by type script hash in dep cells.
+
+However, in many cases, multiple matches do not introduce ambiguity.
+
+There are two different ways to reference the script, via data hash or type hash.
```
xxuejie, line 22: `reference` here is a little vague, it could mean 2 different situations:

* When CKB locates the code for lock/type script to execute
* When an on-chain script tries to find a cell containing certain data

I think this RFC only deals with the first kind, it's better if we be more precise and clarify this.

rfcs/0029-allow-script-multiple-matches-on-identical-code/0029-allow-script-multiple-matches-on-identical-code.md
```diff
@@ -1 +1,93 @@
...
+
+* When the script is referenced by data hash, all the matches are resolved to the identical code.
+* When the script is referenced by type hash, it is possible that the matches can be resolved to different code binaries. The consensus rule can only reject the transaction when this happens.
```
xxuejie, line 25: I would suggest including a real example here.

rfcs/0029-allow-script-multiple-matches-on-identical-code/0029-allow-script-multiple-matches-on-identical-code.md
```diff
@@ -1 +1,93 @@
...
+
+## Specification
+
+When the transaction verifier resolves script code in dep cell via data hash, multiple matches are allowed. This is the same as before.
```
xxuejie, line 29: In the above section, we were talking about `referencing a script`, but here, it's `transaction verifier resolves script code`. There might be the benefit of unifying to a single set of terms, so as not to confuse people. We know they are the same thing, but someone else might need a confirmation.

rfcs/0029-allow-script-multiple-matches-on-identical-code/0029-allow-script-multiple-matches-on-identical-code.md
```diff
@@ -1 +1,93 @@
...
+When the verifier resolves code via type hash, multiple matches are allowed if all the matched cells have the same data, otherwise, the transaction is invalid and the verification fails. This is the modification introduced by this RFC.
+
+## Test Vectors
+
+Examples that fail using the old rule but pass using the new rule.
+
+Multiple matches of data hash.
+
+```
+#  hash(Cell B.data) equals to hash(Cell A.data)
```
xxuejie, line 40: I think this example works in the old rule, correct?

rfcs/0029-allow-script-multiple-matches-on-identical-code/0029-allow-script-multiple-matches-on-identical-code.md
```diff
@@ -1 +1,93 @@
...
+    type:
+      code_hash: hash(Cell A.type)
+      hash_type: Type
+```
+
+## Deployment
+
+The deployment can be performed in two stages.
+
+The first stage will activate the new consensus rule starting from a specific epoch. The mainnet and testnet will use different starting epochs and all other chains will use the new rule from epoch 0.
```
xxuejie, line 87: I think saying `all other chains will use the new rule from epoch 0` is wrong, we can only say new chains instantiated using default settings in CKB after a certain version use the new rule from epoch 0.

doitian:
Assigned reviewers: @xxuejie @quake

xxuejie (approved):

quake (approved):

rfcs/0222-allow-script-multiple-matches-on-identical-code/0222-allow-script-multiple-matches-on-identical-code.md
doitian: Resolved, please review the PR again or simply dismiss the changes request.

janx (approved):

doitian:
FCP will end on 2022-03-28

PR #223, opened 2021-02-04
RFC30: Ensure that index is less than length in input since field using epoch.

doitian:
This document proposes to prevent a bug in the CKB reference implementation and avoid ambiguity. When a transaction input uses the absolute epoch in the `since` field, the `value` is an encoded number `e + i / l`. This RFC proposes to add a rule to verify that `i <= l` or `i` and `l` are both zero.


rfcs/0000-since-absolute-epoch-as-rational-number/0000-since-absolute-epoch-as-rational-number.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Motivation
+
+The `value` part is an encoded number `epoch + index / length`. In CKB, every block is also assigned an epoch number with a fraction. If the block is the `I`th block in the epoch which epoch number is `E` and length is `L`, its epoch number with the fraction is `E + I / L` and `I` must be less than `L`. 
+
+To verify that a transaction is mature since an absolute epoch number with a fraction, the current implementation assumes that `index` must be less than `length` and checks that:
+
+```
+E \> epoch
+or (E == epoch and I / L \>= index/length)
```
yangby-cryptape, line 24:
```suggestion
E > epoch
or (E == epoch and I / L >= index/length)
```

rfcs/0000-since-absolute-epoch-as-rational-number/0000-since-absolute-epoch-as-rational-number.md
```diff
@@ -0,0 +1,87 @@
...
+* The `relative_flag` is absolute (0).
+
+Providing that `value` is the encoded number `epoch + index / length`, the transaction is valid only if `index` is less than `length` or both `index` and `length` are zero \*.
+
+> \* When both `index` and `length` are zero, `value` is converted to `epoch + 0 / 1`. 
+
+If the current epoch number with the fraction is `E + I / L`, the transaction is considered mature when
+
+```
+E + I / L \>= epoch + index / length
```
yangby-cryptape, line 49:
```suggestion
E + I / L >= epoch + index / length
```

rfcs/0000-since-absolute-epoch-as-rational-number/0000-since-absolute-epoch-as-rational-number.md
```diff
@@ -0,0 +1,87 @@
...
+* The `relative_flag` is relative (1).
+
+Providing that `value` is the encoded number `epoch + index / length`, the transaction is valid only if `index` is less than `length` or both `index` and `length` are zero \*.
+
+> \* When both `index` and `length` are zero, `value` is converted to `epoch + 0 / 1`. 
+
+If the cell creation epoch number with fraction is `e + i / l`, and the current epoch number with the fraction is `E + I / L`, the transaction is considered mature when
+
+```
+E + I / L \>= e + i / l + epoch + index / length
```
yangby-cryptape, line 68:
```suggestion
E + I / L >= e + i / l + epoch + index / length
```

yangby-cryptape (commented):
It seems that `>` doesn't require any escape character.

rfcs/0030-ensure-index-less-than-length-in-since/0030-ensure-index-less-than-length-in-since.md
```diff
@@ -1 +1,87 @@
...
+
+```
+E + I / L >= e + i / l + epoch + index / length
+```
+
+using rational number operations.
+
+## Test Vectors
+
+When `since` uses the absolute epoch `99 + 360 / 180`, and the current epoch is `100 + 0 / 180`, the transaction is mature using the old consensus rule but is immature using the new rule.
```
yangby-cryptape, line 75: > but is immature using the new rule.

When using the new rule, it will causes [an error `InvalidSince`](https://github.com/nervosnetwork/ckb/blob/13c7a5bc64c8341655e169007aacfe6c15574707/verification/src/transaction_verifier.rs#L757).

rfcs/0030-ensure-index-less-than-length-in-since/0030-ensure-index-less-than-length-in-since.md
```diff
@@ -1 +1,87 @@
...
+
+The deployment can be performed in two stages.
+
+The first stage will activate the new consensus rule starting from a specific epoch. The mainnet and testnet will use different starting epochs and all other chains will use the new rule from epoch 0.
+
+The second stage is optional. After the new rule has been activated, and the blocks in the chain before activation can also pass the new consensus rule, the old rule can be removed and all chains will use the new rule starting from the genesis block.
+
+## Backward compatibility
+
+The new rule is striker than the old one thus it can be deployed via a soft fork. When most nodes have upgraded to the new version, the old versions can keep up to date although blocks generated by old versions may be rejected by new version nodes.
```
janx, line 87:
```suggestion
The new rule is stricter than the old one thus it can be deployed via a soft fork. When most mining nodes have upgraded to the new version, the old version full nodes can keep up to date. Blocks generated by old version mining nodes may be rejected by new version full nodes. 
```

rfcs/0030-ensure-index-less-than-length-in-since/0030-ensure-index-less-than-length-in-since.md
janx: I suggest making it clear the transaction is invalid not just immature under the new rule.

doitian: My fault. There's two variants of the RFC. We normalize the number in the first version. Later we decide to disallow the malformed fraction number.

doitian, line 75:
```suggestion
When `since` uses the absolute epoch `99 + 360 / 180`, and the current epoch is `100 + 0 / 180`, the transaction is mature using the old consensus rule but is invalid using the new rule.
```

rfcs/0030-ensure-index-less-than-length-in-since/0030-ensure-index-less-than-length-in-since.md
```diff
@@ -1 +1,87 @@
...
+Author: Ian Yang
+Organization: Nervos Foundation
+Created: 2021-02-04
+---
+
+# Ensure That Index Is Less Than Length In the Input Since Field Using Epoch With Fraction
+
+## Abstract
+
+This document proposes to fix a bug in the CKB reference implementation. When a transaction input uses the epoch with fraction in the `since` field, the `value` must ensure that the index is less than the length.
```
xxuejie, line 14: I wouldn't personally say this is a bug in the `CKB reference implementation`. In Lina, a different implementation will also need to replicate this bug for the consensus rules to fully work. In this sense, I would argue this is a bug in the consensus rules, which is derived from the CKB reference implementation.

rfcs/0030-ensure-index-less-than-length-in-since/0030-ensure-index-less-than-length-in-since.md
```diff
@@ -1 +1,87 @@
...
+```
+E > epoch
+or (E == epoch and I / L >= index/length)
+```
+
+This is not consistent when using relative flag in the `since` field, where the `value` is considered as a rational number and added to the cell start epoch.
+
+This document suggests adding a new rule to verify that when `since` uses epoch as the unit, it must ensure that index is less than the length.
+
+See RFC17 [Transaction valid since](../0017-tx-valid-since/0017-tx-valid-since.md) for details of the `since` field encoding patterns.
```
xxuejie, line 31: I suggest we move the link to the very top of this RFC, where we first talked about `value` in `since`. Without necessary information, the whole above section is quite confusing.

rfcs/0030-ensure-index-less-than-length-in-since/0030-ensure-index-less-than-length-in-since.md
```diff
@@ -1 +1,87 @@
...
+
+# Ensure That Index Is Less Than Length In the Input Since Field Using Epoch With Fraction
+
+## Abstract
+
+This document proposes to fix a bug in the CKB reference implementation. When a transaction input uses the epoch with fraction in the `since` field, the `value` must ensure that the index is less than the length.
+
+## Motivation
+
+The `value` part is an encoded number `epoch + index / length`. In CKB, every block is also assigned an epoch number with a fraction. If the block is the `I`th block in the epoch which epoch number is `E` and length is `L`, its epoch number with the fraction is `E + I / L` and `I` must be less than `L`.
```
xxuejie, line 18: From here, it seems to me that `epoch, index and length`, `E, I and L` are all used to represent numbers encoded in `value`.

rfcs/0030-ensure-index-less-than-length-in-since/0030-ensure-index-less-than-length-in-since.md
```diff
@@ -1 +1,87 @@
...
+This document proposes to fix a bug in the CKB reference implementation. When a transaction input uses the epoch with fraction in the `since` field, the `value` must ensure that the index is less than the length.
+
+## Motivation
+
+The `value` part is an encoded number `epoch + index / length`. In CKB, every block is also assigned an epoch number with a fraction. If the block is the `I`th block in the epoch which epoch number is `E` and length is `L`, its epoch number with the fraction is `E + I / L` and `I` must be less than `L`. 
+
+To verify that a transaction is mature since an absolute epoch number with a fraction, the current implementation assumes that `index` must be less than `length` and checks that:
+
+```
+E > epoch
```
xxuejie, line 23: But here, it appears that `E` is used to represent the number encoded in `value`, but `epoch` is used as the actual epoch number for the current cell. I think it's better we clarity both set of terms upfront.

rfcs/0030-ensure-index-less-than-length-in-since/0030-ensure-index-less-than-length-in-since.md
```diff
@@ -1 +1,87 @@
...
+or (E == epoch and I / L >= index/length)
+```
+
+This is not consistent when using relative flag in the `since` field, where the `value` is considered as a rational number and added to the cell start epoch.
```
xxuejie, line 27: Another point: up till this point, we have been talking about `value` in the block header part of CKB. If I understand this correctly, the consensus rule for this part is not changed. We are only changed the validation rule for a similar value used in `since` field represented as epoch number. My suggestion is to clarify this fact, clearly stating what is changed.

Right now what is happening here, is that we spend a huge deal of paragraph documenting what is not changed here, and only using a single sentence below this comment talking about what is changed.

rfcs/0030-ensure-index-less-than-length-in-since/0030-ensure-index-less-than-length-in-since.md
```diff
@@ -1 +1,87 @@
...
+
+This document suggests adding a new rule to verify that when `since` uses epoch as the unit, it must ensure that index is less than the length.
```
xxuejie, line 29: I would recommend either of the following 2 types:

* We remove the above part where we were talking about validation rules in the block header part, where no rules are changed
* We include validation formulas for both the block header, and the since part in transaction input field

In both cases, I also recommend adding real examples since this is the motivation section, not the actual specification below.

rfcs/0030-ensure-index-less-than-length-in-since/0030-ensure-index-less-than-length-in-since.md
```diff
@@ -1 +1,87 @@
...
+## Specification
+
+### Since Absolute Epoch
+
+When an input `since` field is present, and
+
+* The `metric_flag` is epoch (01).
+* The `relative_flag` is absolute (0).
+
+Providing that `value` is the encoded number `epoch + index / length`, the transaction is valid only if `index` is less than `length` or both `index` and `length` are zero \*.
```
xxuejie, line 42: I know this is probably the wrong time to ask for an implementation change, but just for completeness: what is the point of allowing both `index` and `length` as zero? Since this will be converted to `index` as 0 and `length` as 1, what prevents us from removing this rule, and allowing users to fill in `index` as 0, and any arbitrary positive number as `length`?

doitian: This is also derived from the implementation. I keep it unchanged because it's common in some languages to memset the `since` field to zero and only set the epoch part.

rfcs/0030-ensure-index-less-than-length-in-since/0030-ensure-index-less-than-length-in-since.md
```diff
@@ -1 +1,61 @@
...
+
+The deployment can advance in two stages.
+
+The first stage will activate the new consensus rule, starting from a specific epoch. The mainnet and testnet will use different starting epochs and all other chains will use the new rule from epoch 0.
+
+The second stage is optional. After the new rule is active, and the blocks in the chain before activation can also pass the new consensus rule, the old rule is redundant and can be safely removed.
+
+## Backward compatibility
+
+The new rule is striker than the old one, thus it can be deployed via a soft fork. When most nodes have upgraded to the new version, the old versions can keep up to date, although new version nodes may reject blocks generated by old versions.
```
janx, line 61:
```suggestion
The new rule is stricter than the old one thus it can be deployed via a soft fork. When most mining nodes have upgraded to the new version, the old version full nodes can keep up to date. Blocks generated by old version mining nodes may be rejected by new version full nodes.
```

doitian: Applied

xxuejie (approved):

janx (approved):

doitian:
FCP will end on 2022-04-01

PR #224, opened 2021-02-07
RFC31: Add a variable length field in the block header

doitian:
This document proposes adding an optional variable length field to the block header.

### DIFF
--- a/rfcs/0031-variable-length-header-field/0031-variable-length-header-field.md
+++ b/rfcs/0031-variable-length-header-field/0031-variable-length-header-field.md
@@ -95,6 +95,8 @@ For blocks before the activation epoch A, `extension` must be absent. After acti
 
 Pay attention that the `extension` field will occupy the block size. See section [Block and Compact Block Structure](../0020-ckb-consensus-protocol/0020-ckb-consensus-protocol.md#block-and-compact-block-structure) in RFC20 for details.
 
+The uncle blocks packaged in `uncles` will not add the `extension` field.
+
 ### RPC Changes
 
 * The `uncles_hash` is renamed to `extra_hash`.
--- a/rfcs/0031-variable-length-header-field/0031-variable-length-header-field.md
+++ b/rfcs/0031-variable-length-header-field/0031-variable-length-header-field.md
@@ -106,6 +106,239 @@ Pay attention that the `extension` field will occupy the block size.
 2. [Using Molecule Table in New Block Headers](./2-using-molecule-table-in-new-block-headers.md)
 3. [Appending a Hash At the End](./3-appending-a-hash-at-the-end.md)
 
+## Test Vectors
+
+### Block Hash
+
+<details><summary>Block Template</summary>
+
+```json
+{
+  "version": "0x0",
+  "compact_target": "0x20010000",
+  "current_time": "0x17af3f66555",
+  "number": "0x3",
+  "epoch": "0x3e80003000000",
+  "parent_hash": "0xebf229020f333100942279dc33303ae0dfcbe720d8d11818687e6654c157294c",
+  "cycles_limit": "0x2540be400",
+  "bytes_limit": "0x91c08",
+  "uncles_count_limit": "0x2",
+  "uncles": [],
+  "transactions": [
+    {
+      "hash": "0x9110ca9266f89938f09ae6f93cc914b2c856cc842440d56fda6d16ee62543f5c",
+      "required": false,
+      "cycles": "0x19f2d1",
+      "depends": null,
+      "data": {
+        "version": "0x0",
+        "cell_deps": [
+          {
+            "out_point": {
+              "tx_hash": "0xace5ea83c478bb866edf122ff862085789158f5cbff155b7bb5f13058555b708",
+              "index": "0x0"
+            },
+            "dep_type": "dep_group"
+          }
+        ],
+        "header_deps": [],
+        "inputs": [
+          {
+            "since": "0x0",
+            "previous_output": {
+              "tx_hash": "0xa563884b3686078ec7e7677a5f86449b15cf2693f3c1241766c6996f206cc541",
+              "index": "0x7"
+            }
+          }
+        ],
+        "outputs": [
+          {
+            "capacity": "0x2540be400",
+            "lock": {
+              "code_hash": "0x709f3fda12f561cfacf92273c57a98fede188a3f1a59b1f888d113f9cce08649",
+              "hash_type": "data",
+              "args": "0xc8328aabcd9b9e8e64fbc566c4385c3bdeb219d7"
+            },
+            "type": null
+          },
+          {
+            "capacity": "0x2540be400",
+            "lock": {
+              "code_hash": "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8",
+              "hash_type": "type",
+              "args": "0xc8328aabcd9b9e8e64fbc566c4385c3bdeb219d7"
+            },
+            "type": null
+          },
+          {
+            "capacity": "0x2540be400",
+            "lock": {
+              "code_hash": "0x709f3fda12f561cfacf92273c57a98fede188a3f1a59b1f888d113f9cce08649",
+              "hash_type": "data1",
+              "args": "0xc8328aabcd9b9e8e64fbc566c4385c3bdeb219d7"
+            },
+            "type": null
+          }
+        ],
+        "outputs_data": [
+          "0x",
+          "0x",
+          "0x"
+        ],
+        "witnesses": [
+          "0x550000001000000055000000550000004100000070b823564f7d1f814cc135ddd56fd8e8931b3a7040eaf1fb828adae29736a3cb0bc7f65021135b293d10a22da61fcc64f7cb660bf2c3276ad63630dad0b6099001"
+        ]
+      }
+    }
+  ],
+  "proposals": [],
+  "cellbase": {
+    "hash": "0x185d1c46fe3c4a0a1a5ae47203df2aeebbb97ac353abcf2c6a3fc2548ecd4eda",
+    "cycles": null,
+    "data": {
+      "version": "0x0",
+      "cell_deps": [],
+      "header_deps": [],
+      "inputs": [
+        {
+          "since": "0x3",
+          "previous_output": {
+            "tx_hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+            "index": "0xffffffff"
+          }
+        }
+      ],
+      "outputs": [],
+      "outputs_data": [],
+      "witnesses": [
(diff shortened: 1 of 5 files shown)


rfcs/0000-variable-length-header-field/0000-variable-length-header-field.md
```diff
@@ -0,0 +1,154 @@
...
+    header:                 Header,
+    short_ids:              ProposalShortIdVec,
+    prefilled_transactions: IndexTransactionVec,
+    uncles:                 Byte32Vec,
+    proposals:              ProposalShortIdVec,
+    extension_hash:         Byte32,
+}
+```
+
+Before ckb2021 is activated, `extension_hash` must be zero. After activation, the node must verify that `uncles` and `exntesion_hash` match the `extra_hash` in the header.
```
quake, line 107: typo
```suggestion
Before ckb2021 is activated, `extension_hash` must be zero. After activation, the node must verify that `uncles` and `extension_hash` match the `extra_hash` in the header.
```

rfcs/0000-variable-length-header-field/0000-variable-length-header-field.md
```diff
@@ -0,0 +1,154 @@
...
+    transactions:   TransactionVec,
+    proposals:      ProposalShortIdVec,
+    extension_hash: Byte32,
+}
+
+table BlockTransactions {
+    block_hash:     Byte32,
+    transactions:   TransactionVec,
+    uncles:         UncleBlockVec,
+    extension_hash: Byte32,
```
quake, line 94: I think this field can be omitted, because of `CompactBlock` has this field already, peer can use that field to reconstruct the block.

quake (approved):

rfcs/0000-variable-length-header-field/0000-variable-length-header-field.md
```diff
@@ -0,0 +1,143 @@
...
+
+This document proposes a solution to add a variable length field in the block. How the new field is interpreted is beyond the scope of this document and must be defined and deployed via a future soft fork. Although the field is added to the block body, nodes can synchronize block header and this field together in the future version.
+
+## Specification
+
+The block header is encoded as a molecule struct, which consists of fixed length fields. The header binary is just the concatenation of all the fields in sequence.
+
+There are many different ways to add the variable length field to the block header. This RFC proposes to replace the `uncles_hash` in the header with the new field `extra_hash`, which is also a 64-bytes hash. The block will have a new field `extension`.
```
quake, line 30:
```suggestion
There are many different ways to add the variable length field to the block header. This RFC proposes to replace the `uncles_hash` in the header with the new field `extra_hash`, which is also a 32-bytes hash. The block will have a new field `extension`.
```

rfcs/0000-variable-length-header-field/0000-variable-length-header-field.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Specification
+
+The block header is encoded as a molecule struct, which consists of fixed length fields. The header binary is just the concatenation of all the fields in sequence.
+
+There are many different ways to add the variable length field to the block header. This RFC proposes to replace the `uncles_hash` in the header with the new field `extra_hash`, which is also a 32-bytes hash. The block will have a new field `extension`.
+
+The following paragraphs will refer to the current running CKB version as ckb2019, the next fork as ckb2021 and the soft fork introducing the schema of `extension` as ckb2021x.
+
+In CKB ckb2019 and blocks generated in ckb2019, `extra_hash` is the same with the `uncles_hash`. So the block headers serialized by molecule are identical. The version `ckb2019` requires that the `extra_hash` is the hash on `uncles` in the block body. It is all zeros when `uncles` is empty, or `ckbhash` on all the uncle header hashes concatenated together.
```
janx, line 34: ckb2019 or ckb2021? I assume the field is named `uncles_hash` in ckb2019 and renamed to `extra_hash` since ckb2021?

rfcs/0000-variable-length-header-field/0000-variable-length-header-field.md
```diff
@@ -0,0 +1,143 @@
...
+    ```
+    uncles_hash = 0 when uncles is empty, otherwise
+
+    uncles_hash = ckbhash(U1 || U2 || ... || Un)
+      where Ui is the header_hash of the i-th uncle in uncles
+    ```
+
+See Appendix for the default hash function `ckbhash`. The annotation `||` means bytes concatenation.
+
+The version ckb2021 does not verify the content of `extension`, it only requires that its length is at most 96 bytes.
```
janx, line 45: Why 96?

rfcs/0224-variable-length-header-field/0224-variable-length-header-field.md
```diff
@@ -0,0 +1,143 @@
...
+    ```
+    uncles_hash = 0 when uncles is empty, otherwise
+
+    uncles_hash = ckbhash(U1 || U2 || ... || Un)
+      where Ui is the header_hash of the i-th uncle in uncles
+    ```
+
+See Appendix for the default hash function `ckbhash`. The annotation `||` means bytes concatenation.
+
+The version ckb2021 does not verify the content of `extension`, it only requires that its length is at most 96 bytes.
```
yangby-cryptape, line 45: There is another limit: It could be empty but 0 bytes isn't allowed.

rfcs/0000-variable-length-header-field/0000-variable-length-header-field.md
doitian: We'll rename the field for all blocks, including the blocks generated in ckb2019.

rfcs/0000-variable-length-header-field/0000-variable-length-header-field.md
doitian: Explained in the updated version.

yangby-cryptape:
In [RFC 0020](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0020-ckb-consensus-protocol/0020-ckb-consensus-protocol.md#definitions):
> The total size of the first four fields should be no larger than the hard-coded block size limit. The main purpose of implementing a block size limit is to avoid overloading public nodes' bandwidth. The uncle blocks’ proposal zones do not count in the limit as they are usually already synchronized when the block is mined.

Whether the `extension` field count in the block size limit or not?

yangby-cryptape:
The `extension` field is not in `UncleBlock`, I think we should mention that in this RFC.

doitian:
> In [RFC 0020](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0020-ckb-consensus-protocol/0020-ckb-consensus-protocol.md#definitions):
> 
> > The total size of the first four fields should be no larger than the hard-coded block size limit. The main purpose of implementing a block size limit is to avoid overloading public nodes' bandwidth. The uncle blocks’ proposal zones do not count in the limit as they are usually already synchronized when the block is mined.
> 
> Whether the `extension` field count in the block size limit or not?

Included in the block size limit.

yangby-cryptape (approved):

rfcs/0031-variable-length-header-field/0031-variable-length-header-field.md
```diff
@@ -1 +1,371 @@
...
+    proposals:              ProposalShortIdVec,
+    extension:              Bytes,
+}
+```
+
+For blocks before the activation epoch A, `extension` must be absent. After activation, the node must verify that `extension` is absent or a binary with 1 to 96 bytes, and `uncles` and `extension` match the `extra_hash` in the header.
+
+Pay attention that the `extension` field will occupy the block size. See section [Block and Compact Block Structure](../0020-ckb-consensus-protocol/0020-ckb-consensus-protocol.md#block-and-compact-block-structure) in RFC20 for details.
+
+The uncle blocks packaged in `uncles` will not add the `extension` field.
```
janx, line 98:
```suggestion
The uncle blocks packaged in `uncles` will not include the `extension` field.
```

rfcs/0031-variable-length-header-field/0031-variable-length-header-field.md
```diff
@@ -1 +1,371 @@
...
+
+This document proposes a solution to add a variable length field in the block. How the new field is interpreted is beyond the scope of this document and must be defined and deployed via a future soft fork. Although the field is added to the block body, nodes can synchronize the block header and this field together in the future version.
+
+## Specification
+
+The block header is encoded as a molecule struct, which consists of fixed length fields. The header binary is just the concatenation of all the fields in sequence.
+
+There are many different ways to add the variable length field to the block header. This RFC proposes to replace the `uncles_hash` in the header with the new field `extra_hash`, which is also a 32-bytes hash. The block will have a new field `extension`.
```
xxuejie, line 30: `32-byte hash`

rfcs/0031-variable-length-header-field/0031-variable-length-header-field.md
```diff
@@ -1 +1,371 @@
...
+```
+uncles_hash = 0 when uncles is empty, otherwise
+
+uncles_hash = ckbhash(U1 || U2 || ... || Un)
+    where Ui is the header_hash of the i-th uncle in uncles
+```
+
+See Appendix for the default hash function `ckbhash`. The annotation `||` means bytes concatenation.
+
+In blocks generated since epoch A, `extension` can be absent, or any binary with 1 to 96 bytes. The upper limit 96 prevents abusing this field because there's no consensus rule to verify the content of `extension`. The 96 bytes limit allows storing the 64-bytes flyclient hash and extra 32-bytes as a hash on further extension bytes.
```
xxuejie, line 45: Just a note I wasn't sure on this 96 bytes, it says immediately that a PoA based testnet cannot be used for flyclient testing, we might end up in a situation where the remaining 32 bytes become a SMT due to space requirements. But it might be too late to change this value here.

rfcs/0031-variable-length-header-field/0031-variable-length-header-field.md
```diff
@@ -1 +1,371 @@
...
+0xb93dad02d24e9d30c49023d08f84dd8ec34118c1bfec9ed432b75619964686c3
+
+Transaction Hashes:
+0x185d1c46fe3c4a0a1a5ae47203df2aeebbb97ac353abcf2c6a3fc2548ecd4eda
+0x9110ca9266f89938f09ae6f93cc914b2c856cc842440d56fda6d16ee62543f5c
+```
+
+## Appendix
+
+### ckbhash
```
xxuejie, line 346: Can this be a separate RFC? I feel that we might have many RFCs using `ckbhash`.

doitian:
Assigned reviewers: @xxuejie @janx

janx (dismissed):

rfcs/0031-variable-length-header-field/0031-variable-length-header-field.md
doitian: Added an issue for this https://github.com/nervosnetwork/rfcs/issues/320

janx (approved):

xxuejie (approved):

doitian:
FCP will end on 2022-04-01

PR #226, opened 2021-02-26
Revise Since RFC.

doitian:
- Use more precise description.
- Add diagrams and remove codes.
- Fix errors:
    * The timestamp is in the unit milliseconds but not seconds.
    * There are many errors where `<=` is expected but `<` is used or
      vise versa.
    * The Rust code base implementation has many gotchas and must be
      considered as the part of consensus.


rfcs/0017-tx-valid-since/0017-tx-valid-since.md
```diff
@@ -2,335 +2,192 @@
...
+
+    ![](../0027-block-structure/epoch.png)
+
+    * `E` is the epoch number part, the three bytes from the lowest bit 0 to 23.
+    * `I` is the block index in the epoch part, the two bytes from the bit 24 to 39.
+    * `L` is the epoch length part, the two bytes from the bit 40 to 55. There's an **exception** here, `L` is at least 1. If `L` is zero, the whole rational number is considered as `E + 0 / 1`.
+
+**Specially**, when the relative flag is `0` (absolute), the `value` is normalized first. But when the relative flag is `1` (relative), the `value` is interpreted as it is and is added to the creation time directly.
+
+**Absolute Epoch With Fraction Value Normalization**: If the relative flag is `0` (absolute)　and `I` is larger than or equal to `L`, the `value` is normalized to `E + 1 + 0 / 1`.
```
jjyr, line 103: So, by following this rule, we normalize `100 + 201 / 100` to `101 + 0 / 1`. IMO it is a counter-intuitive behaviour, why can't we just reject these values?

jjyr: This phase is outdated #233

doitian: Yes, deprecated after #223 is activated. A notice should be added that the rule is removed after #223 is merged. Before that, thia PR should be reviewed based on the current consensus edition.

Only after #223 is activated in testnet and mainnet, and there are no history blocks in testnet and mainnet violating #223, can we remove it from this RFC.

jjyr (dismissed):

rfcs/0017-tx-valid-since/0017-tx-valid-since.md
```diff
@@ -2,335 +2,192 @@
...
 Status: Proposal
-Author: Jinyang Jiang
+Author: Jinyang Jiang, Ian Yang
 Organization: Nervos Foundation
 Created: 2019-03-11
 ---
 
 # Transaction valid since
 
+<!-- Dialogs are created in LucidChart: https://lucid.app/documents/view/d756089a-2388-4ea4-b61a-3943cbe2620a -->
```
doitian, line 12:
```suggestion
<!-- Diagrams are created in LucidChart: https://lucid.app/documents/view/d756089a-2388-4ea4-b61a-3943cbe2620a -->
```

rfcs/0017-tx-valid-since/0017-tx-valid-since.md
```diff
@@ -2,334 +2,191 @@
...
+* The following two bits choose the time metric used to represent the creation time, destruction time, and the `value`.
+    * `00`: Use block number.
+    * `01`: Use epoch number with fraction.
+    * `10`: Use the timestamp median of previous block headers.
+    * `11`: Invalid. Transaction should not set metric flag to `11`.
+* The next 5 bits are reserved for future extension. They must be set to all zeros now.
+
+### Since Verification
+
+The consensus iterates each input to check whether the cell is mature:
```
janx, line 49:
```suggestion
A CKB node must iterate each input to check whether the cell is mature:
```

Another problem is "the cell is mature" is confusing/improper, "mature" in this RFC is actually an attribute of input not cell.

rfcs/0017-tx-valid-since/0017-tx-valid-since.md
```diff
@@ -2,334 +2,191 @@
...
+
+* The cell is mature when the `since` is 0.
```
janx, line 51:
```suggestion
* An input is always due if its `since` is 0.
```

An example if we use the word "due" with "input".

rfcs/0017-tx-valid-since/0017-tx-valid-since.md
```diff
@@ -2,334 +2,191 @@
...
+* If the relative flag is `0` (absolute), the cell is mature when the destruction time is larger than or equal to the `value`.
```
janx, line 52: I think using "destruction time" to define validation rules is inappropriate, because "destruction time" can only be determined when the transaction is included in a block, but the validation here happens before the transaction is included.

janx (changes requested):

rfcs/0017-tx-valid-since/0017-tx-valid-since.md
```diff
@@ -1,335 +1,152 @@
...
-                self.verify_absolute_lock(since)?;
-                self.verify_relative_lock(since, cell_meta)?;
-            }
-            Ok(())
-        }
-    }
-```
+The new consensus rule allows the transaction input to specify an optional since precondition. CKB nodes must verify the transactions in the commitment zone of a block that all the input since preconditions are fulfilled.
+
+The since precondition locates a unique block in any chain if it is long enough. The precondition is effective before the block, and is fulfilled since the block.
```
janx, line 21:
```suggestion
The since precondition locates a unique block in the past or future on the chain. The precondition is effective before the block, and is fulfilled since the block.
```

rfcs/0017-tx-valid-since/0017-tx-valid-since.md
```diff
@@ -1,335 +1,152 @@
...
+
+There are three metrics to specify the since precondition:
+
+1. via block number,
+2. epoch number with fraction,
+3. or the median timestamp of the preceding 37 blocks.
+
+All of them strictly increase along the block number.
+
+The precondition is absolute or relative. A relative precondition depends on which block has committed the input.
```
janx, line 31: "which block has committed the input" sounds ambiguous - it could mean the block commits the input, or the block commits the cell pointed by the input.

jordanmack:
I will submit an additional pull request with some modifications which will make it easier to read for native English speakers. First I have some questions.

```
When the metric flag is set to epoch number with fraction (01), value represents a rational number E + I / L, where

E has 3 bytes, from the lowest bit 0 to 23.
I has 2 bytes, from the bit 24 to 39.
L has 2 bytes, from the bit 40 to 55.
```

1. It says here that `E` is the "lowest bit 0 to 23". Is this left to right, and then isn't this the highest bit?
2. I see E, I, and L are separate values contained within the `value` field. What is the relevance of `E + I / L`?
3. If a transaction is submitted and the since precondition is not fulfilled, is the transaction always immediately rejected?

@doitian

doitian:
> I will submit an additional pull request with some modifications which will make it easier to read for native English speakers. First I have some questions.
> 
> ```
> When the metric flag is set to epoch number with fraction (01), value represents a rational number E + I / L, where
> 
> E has 3 bytes, from the lowest bit 0 to 23.
> I has 2 bytes, from the bit 24 to 39.
> L has 2 bytes, from the bit 40 to 55.
> ```
> 
> 1. It says here that `E` is the "lowest bit 0 to 23". Is this left to right, and then isn't this the highest bit?

The whole `since` field is an integer, the lowest bit is the least significant bit.

> 2. I see E, I, and L are separate values contained within the `value` field. What is the relevance of `E + I / L`?

The whole `value` part is a rational number for _epoch number with fraction_.

> 3. If a transaction is submitted and the since precondition is not fulfilled, is the transaction always immediately rejected?
> 

Not a requirement of the RFC, but yes, in current node implementation, the tx is rejected immediately.

rfcs/0017-tx-valid-since/0017-tx-valid-since.md
```diff
@@ -1,335 +1,152 @@
...
```
doitian, line 31: How about this?

```suggestion
The precondition is absolute or relative. A relative precondition depends on which block has committed the cell referenced by the input.
```

jordanmack:
> > 2. I see E, I, and L are separate values contained within the `value` field. What is the relevance of `E + I / L`?
> 
> The whole `value` part is a rational number for _epoch number with fraction_.

I still do not understand what this is useful for. Right now is epoch 6083 index 698 length 1028. What is `6083 + 698 / 1028` supposed to be used for?

4. Are all values mentioned always stored in little-endian and read using LSB 0 bit numbering?

doitian:
> > > 2. I see E, I, and L are separate values contained within the `value` field. What is the relevance of `E + I / L`?
> > 
> > 
> > The whole `value` part is a rational number for _epoch number with fraction_.
> 
> I still do not understand what this is useful for. Right now is epoch 6083 index 698 length 1028. What is `6083 + 698 / 1028` supposed to be used for?

Say you set `since` to `absolute, epoch with fraction, 500 + 1 / 2`, you mean that the transaction should commit since the second half of epoch 500.

> 4. Are all values mentioned always stored in little-endian and read using LSB 0 bit numbering?

This RFC does not mention the wire encoding. In block and tx binary format encoded by Molecule, all integers are encoded using little-endian. In RPC Json format, all integers are encoded in hex.

jordanmack:
> Say you set `since` to `absolute, epoch with fraction, 500 + 1 / 2`, you mean that the transaction should commit since the second half of epoch 500.

Is `E + I / L` just a notation being used to describe the epoch in a human-readable format? If yes, is this notation used elsewhere?

doitian:
> > Say you set `since` to `absolute, epoch with fraction, 500 + 1 / 2`, you mean that the transaction should commit since the second half of epoch 500.
> 
> Is `E + I / L` just a notation being used to describe the epoch in a human-readable format? If yes, is this notation used elsewhere?

Yes, it's a notation to explain how three numbers are encoded in a single integer. The `epoch` field in the block header adopts the same encoding method ( https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0027-block-structure/0027-block-structure.md#epoch-uint64 ).

jordanmack:
> Yes, it's a notation to explain how three numbers are encoded in a single integer. The `epoch` field in the block header adopts the same encoding method ( https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0027-block-structure/0027-block-structure.md#epoch-uint64 ).

Sorry for making you repeat yourself, but I need to clarify `E + I / L` to be sure I didn't miss something. I understand this as being a way that humans would communicate with each other about the epoch and fraction.

Example: `500 + 1 / 2` means that the transaction will be allowed to commit in the second half of epoch 500.

If I understand correctly, this has no mathematical usage. This means `500 + 1 / 2 = 500.5` would be of no value to a human or a machine trying to interpret this and we should never try to use the value of `500.5` since it has no meaning and could potentially be an irrational number.

If this is all correct, then I would suggest against using `E + I / L` as a notation because it looks just like a mathematical formula which can be confusing.

The second question I have is about why the `since` value is on the transaction input. I understand this as the `since` value is added to an input on a transaction. The developer who creates the transaction would be the one to optionally specify the `since` value on each input of the transaction. All `since` preconditions on all inputs in the transaction must be fulfilled before the transaction can be committed. Why is it that the `since` value is on each individual input instead of a single `since` value on the transaction itself?

janx:
> The second question I have is about why the since value is on the transaction input. I understand this as the since value is added to an input on a transaction. The developer who creates the transaction would be the one to optionally specify the since value on each input of the transaction. All since preconditions on all inputs in the transaction must be fulfilled before the transaction can be committed. Why is it that the since value is on each individual input instead of a single since value on the transaction itself?

A CKB transaction could be co-authored by multiple users, e.g. in open transactions use case. In that case a user could set conditions on his/her inputs only.

jordanmack:
I have created a new pull request to @doitian's branch with my suggested edits.

- Preview: https://github.com/jordanmack/doitian-rfcs/blob/revise-since-rfc/rfcs/0017-tx-valid-since/0017-tx-valid-since.md
- Pull Request: https://github.com/doitian/rfcs/pull/1
- Commit Details: https://github.com/jordanmack/doitian-rfcs/commit/0a9b202231e69b6e75d468d2c7bd4ade3c6bc419

doitian:
> > Yes, it's a notation to explain how three numbers are encoded in a single integer. The `epoch` field in the block header adopts the same encoding method ( https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0027-block-structure/0027-block-structure.md#epoch-uint64 ).
> 
> Sorry for making you repeat yourself, but I need to clarify `E + I / L` to be sure I didn't miss something. I understand this as being a way that humans would communicate with each other about the epoch and fraction.
> 
> Example: `500 + 1 / 2` means that the transaction will be allowed to commit in the second half of epoch 500.
> 
> If I understand correctly, this has no mathematical usage. This means `500 + 1 / 2 = 500.5` would be of no value to a human or a machine trying to interpret this and we should never try to use the value of `500.5` since it has no meaning and could potentially be an irrational number.
> 
> If this is all correct, then I would suggest against using `E + I / L` as a notation because it looks just like a mathematical formula which can be confusing.

What's your suggestion? Using rational number notation can simplify the explanation because adding and comparison works for rational numbers just like integers.

> The second question I have is about why the `since` value is on the transaction input. I understand this as the `since` value is added to an input on a transaction. The developer who creates the transaction would be the one to optionally specify the `since` value on each input of the transaction. All `since` preconditions on all inputs in the transaction must be fulfilled before the transaction can be committed. Why is it that the `since` value is on each individual input instead of a single `since` value on the transaction itself?

Contracts may have conflict requirement on the since. For example, contract requires the since is at least 100 blocks, and another contract may require the since at least 1 epoch.

jordanmack:
> What's your suggestion? Using rational number notation can simplify the explanation because adding and comparison works for rational numbers just like integers.

The term "rational number" is indicating a single number. The field has three values that can be assembled and referred to as a single "rational number" if you want a human readable format. But these three numbers are encoded into a u64, which is also a single number.

The notation `E + I / L` looks like a formula to calculate something, but it is not a formula. The terminology is not incorrect, but I believe it is less confusing to explain it as three numbers then describe the epoch as a fraction that is written as `E I/L`. The change is subtle, but represented as fraction without the "+" makes it less likely to be mistaken for a formula.

jordanmack (approved):

doitian:
On Sun, Aug 28, 2022, 8:32 AM Jordan Mack ***@***.***> wrote:

> I will submit an additional pull request with some modifications which
> will make it easier to read for native English speakers. First I have some
> questions.
>
> When the metric flag is set to epoch number with fraction (01), value represents a rational number E + I / L, where
>
> E has 3 bytes, from the lowest bit 0 to 23.
> I has 2 bytes, from the bit 24 to 39.
> L has 2 bytes, from the bit 40 to 55.
>
>
>    1. It says here that E is the "lowest bit 0 to 23". Is this left to
>    right, and then isn't this the highest bit?
>
>
The whole since field is an integer. The lowest bits are the least
significant bits.


>    1. I see E, I, and L are separate values contained within the value
>    field. What is the relevance of E + I / L?
>
>
The whole is a rational number.


>    1. If a transaction is submitted and the since precondition is not
>    fulfilled, is the transaction always immediately rejected?
>
>
It is not a requirement from the RFC, but yes, in the current node
implementation, the tx is rejected immediately.

doitian:
The email took half a month to arrive at GitHub 😂

PR #235, opened 2021-05-14
RFC: propose to use the GitHub PR number as the RFC number

doitian:


doitian:
https://github.com/nervosnetwork/rfcs/pull/222#pullrequestreview-713339871

PR #237, opened 2021-05-27
RFC34: vm syscalls 2

mohanson:


rfcs/0000-vm-syscalls-2/0000-vm-syscalls-2.md
```diff
@@ -0,0 +1,87 @@
+---
+Number: "0000"
```
doitian, line 2: Use "0237"

rfcs/0237-vm-syscalls-2/0237-vm-syscalls-2.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+*Current Cycles* returns current cycle consumption just before executing this syscall. This syscall consumes 500 cycles.
+
+
+### Exec
+[exec]: #exec
+
+Exec runs an executable file from specified cell data in the context of an already existing machine, replacing the previous executable. The used cycles does not change, but the code, registers and memory of the vm are replaced by those of the new program. It's cycles consumption consists of two parts:
+
```
sking789, line 53: Will the code, register and memory be restored to the previous executable program, after the specified new executable file execution finished?

mohanson: No

rfcs/0237-vm-syscalls-2/0237-vm-syscalls-2.md
```diff
@@ -0,0 +1,87 @@
...
+- Fixed 500 cycles
+- Initial Loading Cycles [1]
+
+*Exec* syscall has a signature like following:
+
+```c
+int ckb_exec(size_t index, size_t source, size_t place, size_t bounds, int argc, char* argv[])
```
huwenchao, line 60: 1. Since the code, registers and memory of the vm are replaced when executing, is this function supposed to be used like pure function? It can not change any state of current script, right?
2. How do I pass something back to current script in `ckb_exec`? E.g. I developed a binary to calculate hash, I can accept contents with argc and argv, but how do I pass the hash back?
3. `argv` can not accept bytes with `\0` inside, is it better to pass arguments with `size_t arg_length, char arg[]`?
It seems that the `run` function in ckb-vm supports that. https://github.com/nervosnetwork/ckb-vm/blob/develop/src/lib.rs#L36

mohanson: > 1. Since the code, registers and memory of the vm are replaced when executing, is this function supposed to be used like pure function? It can not change any state of current script, right?
> 2. How do I pass something back to current script in ckb_exec? E.g. I developed a binary to calculate hash, I can accept contents with argc and argv, but how do I pass the hash back?

`ckb_exec()` behaviors like Linux's `exec()` https://man7.org/linux/man-pages/man3/exec.3.html, it's a diverging functions:

- When function call succeed, it never returns
- When function call failed, it returns an error code

xxuejie: 1. `exec` is modeled after real `exec` syscalls on Linux: https://man7.org/linux/man-pages/man3/exec.3.html, we are talking about a new program being brought in and executed. Current running script can be considered discarded and totally replaced after the `exec` call.
2. As mentioned in `exec`, this new model is designed at a program level, not a library level. Once you pass control to the new binary, there is no way going back. In the future we might bring in `fork` which combined with `exec`, can perform the job of running a binary and return, but for this version, we want to start small here. Most use cases we have now, is to dynamically decide a signature verification algorithm to use, `exec` alone already perfectly suits this case
3. I do admit this is a problem, but I cannot think of a perfect solution that works in C ABI. Suggestions are welcome.

doitian:
Please mention #232 how the node selects VM versions and these syscalls are only available since ckb version 1 and ckb2021.

```
[rfc232](../0232-ckb-vm-version-selection/0232-ckb-vm-version-selection.md)
```

mohanson:
> Please mention #232 how the node selects VM versions and these syscalls are only available since ckb version 1 and ckb2021.
> 
> ```
> [rfc232](../0232-ckb-vm-version-selection/0232-ckb-vm-version-selection.md)
> ```

done

rfcs/0237-vm-syscalls-2/0237-vm-syscalls-2.md
```diff
@@ -0,0 +1,89 @@
...
+As shown above, *VM Version* syscall has a signature like following:
+
+```c
+int ckb_vm_version()
+{
+  return syscall(2041, 0, 0, 0, 0, 0, 0);
+}
+```
+
+*VM version* syscall returns current running VM version, so far 2 values will be returned: 0 for Lina CKB-VM version, 1 for the new hardfork version. This syscall consumes 500 cycles.
```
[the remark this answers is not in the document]
mohanson: Yes, this is an obvious error.

doitian:
Please change RFC numbers according to https://github.com/nervosnetwork/rfcs/issues/246 . Sorry for the inconvenience.

doitian:
Assigned reviewers: @quake @doitian

doitian (approved):

quake (approved):

doitian:
FCP will end on 2022-03-28

PR #252, opened 2021-08-11
ci: add codespell

doitian:
- fix spell errors
- add workflow to check spell

### DIFF
--- a/.github/workflows/spell.yaml
+++ b/.github/workflows/spell.yaml
@@ -1,0 +1,26 @@
+name: Spelling Check
+
+on: [push, pull_request]
+
+jobs:
+  spellcheck:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+      with:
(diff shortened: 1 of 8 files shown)


.github/workflows/spell.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
+    - uses: actions/cache@v2
+      name: Configure pip caching
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-spellcheck
+        restore-keys: |
+          ${{ runner.os }}-pip-spellcheck
```
keroro520, line 19: Trying executing `pip install codespell` inside my docker is fast. So why use actions/cache here?

keroro520 (changes requested):

.github/workflows/spell.yaml
doitian: copy from somewhere

zhangsoledad (approved):

PR #271, opened 2021-11-18
More status

janx:
Revised RFC process and added more status to both tracks.

Based on https://github.com/nervosnetwork/rfcs/pull/270

PR #274, opened 2021-11-26
Revert "More status (#271)"

malinoisls01:
This reverts commit ff7012938ca23f373888bf82477c295e3de5d494.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,9 +21,9 @@ The RFC process attempts to be as simple as possible at beginning and evolves wi
 
 ### 1. Discuss Your Idea with Community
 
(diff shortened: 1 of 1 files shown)


malinoisls01:
Olympus Pro - is a form of Liquidity Mining (LM) that helps to stabilize price while allowing growth in token circulation through a LM that sells the BRIGHT token at a discount while accumulating ‘whitelisted tokens and/or LPs’ the dao wants to accumulate assets in it’s reserve. The allocated LM tokens that would normally be used for farming which often results in direct sell pressure. Instead Olympus Pro uses what would be normally used for farm rewards to help offset the cost of the BRIGHT token to allow buyers to buy the token at a discount. Secondly, the bonding mechanism sells the BRIGHT token when the token value is greater than the backed assets to help stabilize price while accumulating a reserve.

PR #282, opened 2021-12-21
finalize 006

kilb:


janx (changes requested):
As this is a RFC on `Standards Track`, status should be `Active` instead of `Final`.

janx (approved):

zhangsoledad (approved):

Issue #246, opened 2021-07-23
Trace RFC Number Allocation


doitian:
| Number | PR |
| -------- | --- |
| 28 | Next Available Number |

janx:
PR#333 is closed and replaced by https://github.com/nervosnetwork/rfcs/pull/335

XuJiandong:
reserve RFC number (42) for https://github.com/nervosnetwork/rfcs/pull/343

zhangsoledad:
reserve RFC number (48) for https://github.com/nervosnetwork/rfcs/pull/416
reserve RFC number (49) for https://github.com/nervosnetwork/rfcs/pull/419
reserve RFC number (50) for https://github.com/nervosnetwork/rfcs/pull/418
reserve RFC number (51) for https://github.com/nervosnetwork/rfcs/pull/417

XuJiandong:
reserve RFC number (42) for Extensible UDT(xUDT)

zhangsoledad:
reserve RFC number (53) for https://github.com/nervosnetwork/rfcs/pull/442

mohanson:
reserve RFC number (54) for https://github.com/nervosnetwork/rfcs/pull/451

PR #308, opened 2022-01-27
rfc(0039): Add cheque lock rfc

duanyytop:


rfcs/0038-cheque/0038-cheque.md
```diff
@@ -0,0 +1,324 @@
...
+
+1. If a signature is provided in witness, the lock continues with the cheque logic below:
+
+   - 1.a. If the provided signature in witness fails validation with the receiver or the sender secp256k1 public key hash, the cheque lock returns withe an error state.
+
+   - 1.b. If the provided signature in witness can be validated with the receiver secp256k1 public key hash, the cheque lock continues with the claim logic below:
+
+     - 1.b.i. It loops through all input cells using the current cheque lock script(notice here the lock script we refer to include public key hash, meaning if a transaction contains 2 cells using the same cheque lock code, but different public key hash, they will be treated as different lock script, and each will perform the script unlock rule checking independently), if the since of 2 inputs are not zero, the cheque lock returns with an error state
+
+     - 1.b.ii. It loops through all output cells using the receiver lock hash, if the sum of the output cells capacity is not equal to the sum of the cheque input cells capacity, the cheque lock returns with an error state
```
janx, line 46: Using the receiver lock hash or sender lock hash?

rfcs/0038-cheque/0038-cheque.md
```diff
@@ -0,0 +1,324 @@
...
+    Normal Cell:
+        Capacity: 165 CKBytes
+        Lock: <sender_secp256k1_blake2b_lock_script>
+
+ Witnesses :
+      < valid signature for another public key hash >
+      < valid signature for receiver public key hash >
+```
+
+When a signature is provided and can be validated by the receiver public key hash, and the sum of sender output cells capacity is equal to the sum of the cheque input cells capacity, the cheque cells can be unlocked. In this example a cheque cell is converted back to a sender normal cell and the SUDT is transferred from the sender to an arbitrary lock script set by the receiver.
```
janx, line 136: "empty cell" might be a better term than "normal cell"?

rfcs/0038-cheque/0038-cheque.md
```diff
@@ -0,0 +1,324 @@
...
+        Capacity: 165 CKBytes
+        Lock: <sender_secp256k1_blake2b_lock_script>
+
+Witnesses:
+    < 0x >
+    < valid signature for receiver_receiver public key hash >
+
+```
+
+Here the transaction inputs contain a receiver secp256k1_blake160 cell whose first 20 bytes of lock script hash is equal to the `receiver_secp256k1_blake2b_lock_hash[0..20]` of the cheque cell lock args, and the signature can be validated by the receiver public key hash, and the sum of sender output cells capacity is equal to the sum of the cheque input cells capacity, the cheque cell can be unlocked.
```
janx, line 174:
```suggestion
In this example, the transaction inputs include a receiver secp256k1_blake160 cell whose first 20 bytes of lock script hash is equal to the `receiver_secp256k1_blake2b_lock_hash[0..20]` of the cheque cell lock args, and the signature can be validated by the receiver public key hash, and the sum of sender output cells capacity is equal to the sum of the cheque input cells capacity, the cheque cell can be unlocked.
```

The witness layout in this scenario is not defined in RFC.

rfcs/0038-cheque/0038-cheque.md
```diff
@@ -0,0 +1,324 @@
...
+    Normal Cell:
+        Capacity: 199.99 CKBytes
+        Lock: <sender_secp256k1_blake2b_lock_script>
+
+Witnesses:
+    < 0x >
+    < valid signature for sender public key hash >
+```
+
+Here the transaction inputs contain a sender secp256k1_blake160 cell whose first 20 bytes of lock script hash is equal to the `sender_secp256k1_blake2b_lock_hash[0..20]` of the cheque cell lock args, and the signature can be validated by the sender public key hash, and the since of the cheque input cell is same as `0xA000000000000006` which means the tx failed verification unless it is 6 epochs later since the input cells get confirmed on-chain, the cheque cell can be unlocked. In this example a cheque cell is converted back to a sender normal cell with SUDT.
```
janx, line 239:
```suggestion
In this example, the transaction inputs include a sender secp256k1_blake160 cell whose first 20 bytes of lock script hash is equal to the `sender_secp256k1_blake2b_lock_hash[0..20]` of the cheque cell lock args, and the signature can be validated by the sender public key hash, and the since of the cheque input cell is same as `0xA000000000000006` which means the tx failed verification unless it is 6 epochs later since the input cells get confirmed on-chain, the cheque cell can be unlocked. In this example a cheque cell is converted back to a sender normal cell with SUDT.
```

rfcs/0038-cheque/0038-cheque.md
```diff
@@ -0,0 +1,324 @@
...
+
+## Deployment
+
+[cheque](https://talk.nervos.org/t/sudt-cheque-deposit-design-and-implementation/5209) ([Source Code](https://github.com/nervosnetwork/ckb-cheque-script/tree/4ca3e62ae39c32cfcc061905515a2856cad03fd8)) allows a sender to temporarily provide cell capacity in asset transfer.
```
janx, line 243: I think the link to the original proposal and source code can be moved into a Reference section at the end.

rfcs/0038-cheque/0038-cheque.md
```diff
@@ -0,0 +1,324 @@
...
+and the `out_point` of `cheque` is
+
+```
+{
+  tx_hash: 0x1b16769dc508c8349803fe65558f49aa8cf04ca495fbead42513e69e46608b6c,
+  index: 0x0
+}
+```
+
+## Notes
```
janx, line 309:
```suggestion
## Reproducible Build
```

rfcs/0038-cheque/0038-cheque.md
```diff
@@ -0,0 +1,324 @@
...
+
+Reproducible build is supported to verify the deploy script. To build the deployed cheque lock script above, one can use the following steps:
```
janx, line 311:
```suggestion
Anyone can use the reproducible build to verify that the deployed script is consistent with the source code. To build the deployed cheque lock script, one can use the following steps:
```

janx (changes requested):

rfcs/0038-cheque/0038-cheque.md
```diff
@@ -0,0 +1,324 @@
...
+type:
+    <simple_udt_type_script>
+lock:
+    code_hash: cheque lock script
+    args: <20 byte receiver secp256k1-blake2b-sighash-all lock hash> <20 byte sender secp256k1-blake2b-sighash-all lock hash>
+```
+
+When the sender wants to transfer the SUDT to another, he/she needs to create a cheque cell whose lock args includes the receiver’s and sender’s secp256k1-blake2b-sighash-all lock hash. The receiver can unlock the related cheque cell to get the SUDT and give back the CKB to the sender.
+
+When the cheque cell exceeds the lock-up period(6 epochs), the sender can withdraw his/her assets.
```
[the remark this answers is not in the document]
duanyytop: Yes

rfcs/0038-cheque/0038-cheque.md
```diff
@@ -0,0 +1,324 @@
...
+
+     - 1.c.i. It loops through all input cells using the current cheque lock script, if the any since of the cheque input cells is not same as `0xA000000000000006` which means the tx failed verification unless it is 6 epochs later since the input cells get confirmed on-chain, the cheque lock returns with an error state
+
+2. If a signature is not provided in witness, the lock continues with the cheque logic below:
+
+   - 2.a. It loops through all input cells using the receiver and the sender lock hash, if no matching input cells are found, the cheque lock returns with an error state
+
+   - 2.b. If the matching input cells with the receiver lock hash are found, it does the same work as the 1.b.i and 1.b.ii and returns the same error state
+
+     - 2.b.iii. It loops through all input cells using the receiver lock hash, if the matching inputs are not found or the first related witness is empty(the witness is not WitnessArgs or the lock of WitnessArgs is empty), the cheque lock returns with an error state
```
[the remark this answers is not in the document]
duanyytop: > Here `2.b.iii` should be `2.b.i`?

`2.b.iii` takes over `1.b.i` and `1.b.ii` and maybe `2.b.i` is more clear.

rfcs/0038-cheque/0038-cheque.md
duanyytop: It should be the sender lock hash, this rule will ensure that the sender's CKB will not be lost

rfcs/0038-cheque/0038-cheque.md
```diff
@@ -0,0 +1,332 @@
...
+        Type: <sudt_type_script>
+        Data:
+            sudt_amount: 300 UDT
+    Empty Cell:
+        Capacity: 165 CKBytes
+        Lock: <sender_secp256k1_blake2b_lock_script>
+
+ Witnesses :
+      < valid signature for another public key hash >
+      < valid signature for receiver public key hash >
```
TheWaWaR, line 135: I think the witnesses should be(the order is not rignt?):
```
Witnesses :
      < valid signature for receiver public key hash >
      < valid signature for other public key hash >
```

rfcs/0038-cheque/0038-cheque.md
```diff
@@ -0,0 +1,332 @@
...
+            sudt_amount: 200 UDT
+
+Outputs:
+    SUDT Cell:
+        Capacity: 199.99 CKBytes
+        Lock: <receiver_secp256k1_blake2b_lock_script>
+        Type: <sudt_type_script>
+        Data:
+            sudt_amount: 300 UDT
+    Empty Cell:
```
TheWaWaR, line 166: I think this not necessary be `Empty Cell`, use `Any Cell` would be better?

rfcs/0038-cheque/0038-cheque.md
duanyytop: Updated.

doitian (dismissed):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nervosnetwork/rfcs/308)* status: 0 of 1 files reviewed, 28 unresolved discussions (waiting on @doitian, @duanyytop, @janx, and @TheWaWaR)

___
*[rfcs/0038-cheque/0038-cheque.md, line 2 at r5](https://reviewable.io/reviews/nervosnetwork/rfcs/308#-MvacHsJ3ONeZa7O7pyV:-MvacHsJ3ONeZa7O7pyW:b-rtchnm) ([raw file](https://github.com/nervosnetwork/rfcs/blob/c7480941049c057490a815024e47541481b40b8b/rfcs/0038-cheque/0038-cheque.md#L2)):*
> ```Markdown
> ---
> Number: "0038"
> ```

38 has been reserved, please use 39.

Check https://github.com/nervosnetwork/rfcs/issues/246 for details.

rfcs/0038-cheque/0038-cheque.md
duanyytop: @janx any suggestions?

rfcs/0039-cheque/0039-cheque.md
```diff
@@ -0,0 +1,333 @@
...
+
+Here we try to solve the problem by introducing a new cheque lock script, which can be claimed not only by the receiver who does not have an ACP cell to receive the SUDT, but also withdrawing by the sender after lock-up period(6 epochs). When the sender wants to send the SUDT to a receiver who does not have an ACP cell for the SUDT, he/she can transfer the SUDT to a cheque cell and then wait for the receiver to claim it. The receiver can get the cheque cell information from the chain and then decide whether to create a cell to receive the SUDT so that the receiver can receive the SUDT without creating an ACP cell in advance and the sender doesn’t need to pay the additional costs.
+
+### Cheque Script Structure
+
+A cheque cell is the cell whose lock script is cheque script and its structure looks like following:
```
janx, line 18: Is a cell with empty data, empty type, and cheque script as lock considered a cheque cell?

rfcs/0039-cheque/0039-cheque.md
```diff
@@ -0,0 +1,333 @@
...
+$ cd .. && make all-via-docker
+$ cd ../../.. && capsule build --release
+```
+
+A draft of this specification has already been released, reviewed, and discussed in the community at [here](https://talk.nervos.org/t/sudt-cheque-deposit-design-and-implementation/5209) for quite some time.
+
+## References
+
+[1] SUDT Cheque Deposit Design and Implementation, https://talk.nervos.org/t/sudt-cheque-deposit-design-and-implementation/5209
+
```
janx, line 332: Redundant newline.
```suggestion
```

rfcs/0038-cheque/0038-cheque.md
janx: I think `Empty Cell` is appropriate for this example because that's what is demonstrated. `Any Cell` sounds good too. Maybe it's even better to add a comment to make clear that any cell can be used here.

rfcs/0039-cheque/0039-cheque.md
duanyytop: Yes

rfcs/0039-cheque/0039-cheque.md
```diff
@@ -0,0 +1,332 @@
...
+A cheque cell is the cell whose lock script is cheque script and its structure looks like following:
+
+```
+data:
+    amount: uint128
+type:
+    <simple_udt_type_script>
+lock:
+    code_hash: cheque lock script
+    args: <20 byte receiver secp256k1-blake2b-sighash-all lock hash> <20 byte sender secp256k1-blake2b-sighash-all lock hash>
```
quake, line 27: I think the receiver and sender's lock can be arbitrary lock:

```suggestion
    args: <20 byte receiver lock hash> <20 byte sender lock hash>
```

duanyytop: The signature in witness can only be validated with `secp256k1-blake2b-sighash-all` algorithm

WhiteR4bbitt:
Hey, I had a question about this Lock and I guess varying on the answer I wanted to know if my "rewrite" of the first couple of paragraphs was fitting.

Is it safe to assume this acts as the `Lock` and anyone could use any `Type` + `Data` ?

If yes, it seems more broad than just SUDTs as per this RFC now, which is why I did attempt rewriting this a bit:

---

This RFC describes a new **Lock Script for CKB** that creates a new Cell allowing *anyone to lock anything for anyone*, whether an *NFT*, a [Simple UDT](../0025-simple-udt/0025-simple-udt.md) or something yet discovered. This allows user interactions to be more seamless if the receiver does not have an available Cell (such as an **[Anyone-Can-Pay Lock](../0026-anyone-can-pay/0026-anyone-can-pay.md)ed Cell**). The receiver needs a Cell compatible with what the sender is sending and the problem is the receiver may not know in advance that someone is going to send them something, such as an *SUDT*. For now the receiver can only be notified via an off-chain notification that a Cell is needed. Or the sender needs to create the Cell (ACP) for the receiver containing the SUDT. These two ways will cause a problem that the receiver cannot receive the SUDT conveniently or the sender needs to pay additional costs.

Introducing the ***Cheque** Lock Script*, which can be claimed by a receiver who does not have an available Cell, but can still be withdrawn if needed by the sender after a lock-up period(6 epochs). The sender creates a new Cell **locked by the Cheque script** and waits for the receiver to claim it. The receiver can get the Cell information from the chain and then decide whether to create a new Cell and *claim the data* (NFT, SUDT, etc). This allows the receiver to receive any kind of pre-made Cell without knowing in advance and additionally the sender doesn’t need to pay the additional costs. Once the Cell has been claimed the sender can reclaim any capacity from the now Empty Cell.

---

Raw MD:
```
This RFC describes a new **Lock Script for CKB** that creates a new Cell allowing *anyone to lock anything for anyone*, whether an *NFT*, a [Simple UDT](../0025-simple-udt/0025-simple-udt.md) or something yet discovered. This allows user interactions to be more seamless if the receiver does not have an available Cell (such as an **[Anyone-Can-Pay Lock](../0026-anyone-can-pay/0026-anyone-can-pay.md)ed Cell**). The receiver needs a Cell compatible with what the sender is sending and the problem is the receiver may not know in advance that someone is going to send them something, such as an *SUDT*. For now the receiver can only be notified via an off-chain notification that a Cell is needed. Or the sender needs to create the Cell (ACP) for the receiver containing the SUDT. These two ways will cause a problem that the receiver cannot receive the SUDT conveniently or the sender needs to pay additional costs.

Introducing the ***Cheque** Lock Script*, which can be claimed by a receiver who does not have an available Cell, but can still be withdrawn if needed by the sender after a lock-up period(6 epochs). The sender creates a new Cell **locked by the Cheque script** and waits for the receiver to claim it. The receiver can get the Cell information from the chain and then decide whether to create a new Cell and *claim the data* (NFT, SUDT, etc). This allows the receiver to receive any kind of pre-made Cell without knowing in advance and additionally the sender doesn’t need to pay the additional costs. Once the Cell has been claimed the sender can reclaim any capacity from the now Empty Cell.
```
 
I think the examples could be re-written given this as well or just a few added showcasing a different "use case".  

Again this is all under the assumption we can use this LOCK with ANY TYPE/DATA not just SUDT.

This just seems interesting because you could have it hold NFTs for users, kind of acting as a layer of "claimed". 
ex: I want to buy an NFT off X Marketplace, but I actually want it for Bob, so I would have to buy the NFT and then could create this Lock to actually give it to Bob, but perhaps Bob didn't want it after all, I can reclaim the NFT and do as I wish afterwards.

A Marketplace could even have the "Gift to" built in with this, but this allows a wall of protection against just spamming items to users. (I believe OpenSea had an issue with this kind of feature recently..).
  
*P.S. If I can throw a name change in the ring, if its not too late? **"Holding"** Lock instead of "Cheque"* Lock

rfcs/0039-cheque/0039-cheque.md
```diff
@@ -0,0 +1,332 @@
...
+Category: Standards Track
+Status: Proposal
+Author: Dylan Duan
+Organization: Nervina Labs
+Created: 2022-01-27
+---
+
+# Cheque Lock
+
+This RFC describes a new lock script for CKB that can send SUDT([Simple UDT](../0025-simple-udt/0025-simple-udt.md)) to anyone that does not have an ACP([any-one-pay](../0026-anyone-can-pay/0025-anyone-can-pay.md)) cell for the SUDT. Previously, the receiver needs to create an ACP cell for the SUDT, and the receiver cannot actually know in advance that someone is going to send a new SUDT to him/her, for now the receiver can only be notified via an off-chain notification that an ACP cell is needed. Or the sender needs to give the receiver an ACP cell of the SUDT when sending the SUDT. These two ways will cause a problem that the receiver cannot receive the SUDT conveniently or the sender needs to pay additional costs.
```
WhiteR4bbitt, line 12:
```suggestion
This RFC describes a new lock script for CKB that can send SUDT([Simple UDT](../0025-simple-udt/0025-simple-udt.md)) to anyone that does not have an ACP([any-one-pay](../0026-anyone-can-pay/0026-anyone-can-pay.md)) cell for the SUDT. Previously, the receiver needs to create an ACP cell for the SUDT, and the receiver cannot actually know in advance that someone is going to send a new SUDT to him/her, for now the receiver can only be notified via an off-chain notification that an ACP cell is needed. Or the sender needs to give the receiver an ACP cell of the SUDT when sending the SUDT. These two ways will cause a problem that the receiver cannot receive the SUDT conveniently or the sender needs to pay additional costs.
```

I did leave a comment, but outside of that I noticed the typo in this URL

duanyytop:
> Hey, I had a question about this Lock and I guess varying on the answer I wanted to know if my "rewrite" of the first couple of paragraphs was fitting.
> 
> Is it safe to assume this acts as the `Lock` and anyone could use any `Type` + `Data` ?
> 
> If yes, it seems more broad than just SUDTs as per this RFC now, which is why I did attempt rewriting this a bit:
> 
> This RFC describes a new **Lock Script for CKB** that creates a new Cell allowing _anyone to lock anything for anyone_, whether an _NFT_, a [Simple UDT](../0025-simple-udt/0025-simple-udt.md) or something yet discovered. This allows user interactions to be more seamless if the receiver does not have an available Cell (such as an **[Anyone-Can-Pay Lock](../0026-anyone-can-pay/0026-anyone-can-pay.md)ed Cell**). The receiver needs a Cell compatible with what the sender is sending and the problem is the receiver may not know in advance that someone is going to send them something, such as an _SUDT_. For now the receiver can only be notified via an off-chain notification that a Cell is needed. Or the sender needs to create the Cell (ACP) for the receiver containing the SUDT. These two ways will cause a problem that the receiver cannot receive the SUDT conveniently or the sender needs to pay additional costs.
> 
> Introducing the _**Cheque** Lock Script_, which can be claimed by a receiver who does not have an available Cell, but can still be withdrawn if needed by the sender after a lock-up period(6 epochs). The sender creates a new Cell **locked by the Cheque script** and waits for the receiver to claim it. The receiver can get the Cell information from the chain and then decide whether to create a new Cell and _claim the data_ (NFT, SUDT, etc). This allows the receiver to receive any kind of pre-made Cell without knowing in advance and additionally the sender doesn’t need to pay the additional costs. Once the Cell has been claimed the sender can reclaim any capacity from the now Empty Cell.
> 
> Raw MD:
> 
> ```
> This RFC describes a new **Lock Script for CKB** that creates a new Cell allowing *anyone to lock anything for anyone*, whether an *NFT*, a [Simple UDT](../0025-simple-udt/0025-simple-udt.md) or something yet discovered. This allows user interactions to be more seamless if the receiver does not have an available Cell (such as an **[Anyone-Can-Pay Lock](../0026-anyone-can-pay/0026-anyone-can-pay.md)ed Cell**). The receiver needs a Cell compatible with what the sender is sending and the problem is the receiver may not know in advance that someone is going to send them something, such as an *SUDT*. For now the receiver can only be notified via an off-chain notification that a Cell is needed. Or the sender needs to create the Cell (ACP) for the receiver containing the SUDT. These two ways will cause a problem that the receiver cannot receive the SUDT conveniently or the sender needs to pay additional costs.
> 
> Introducing the ***Cheque** Lock Script*, which can be claimed by a receiver who does not have an available Cell, but can still be withdrawn if needed by the sender after a lock-up period(6 epochs). The sender creates a new Cell **locked by the Cheque script** and waits for the receiver to claim it. The receiver can get the Cell information from the chain and then decide whether to create a new Cell and *claim the data* (NFT, SUDT, etc). This allows the receiver to receive any kind of pre-made Cell without knowing in advance and additionally the sender doesn’t need to pay the additional costs. Once the Cell has been claimed the sender can reclaim any capacity from the now Empty Cell.
> ```
> 
> I think the examples could be re-written given this as well or just a few added showcasing a different "use case".
> 
> Again this is all under the assumption we can use this LOCK with ANY TYPE/DATA not just SUDT.
> 
> This just seems interesting because you could have it hold NFTs for users, kind of acting as a layer of "claimed". ex: I want to buy an NFT off X Marketplace, but I actually want it for Bob, so I would have to buy the NFT and then could create this Lock to actually give it to Bob, but perhaps Bob didn't want it after all, I can reclaim the NFT and do as I wish afterwards.
> 
> A Marketplace could even have the "Gift to" built in with this, but this allows a wall of protection against just spamming items to users. (I believe OpenSea had an issue with this kind of feature recently..).
> 
> _P.S. If I can throw a name change in the ring, if its not too late? **"Holding"** Lock instead of "Cheque"_ Lock

Your suggestions really make sense for us and we will update the RFC later. Thanks~

duanyytop:
@WhiteR4bbitt I updated the RPC with the commit: https://github.com/nervosnetwork/rfcs/pull/308/commits/345b19e0f9d3b245b983ff38106506f7bc90ac5e.

WhiteR4bbitt:
> * I used SUDT as an example to refer to custom assets for the convenience of description
Yeah that makes sense, the re-wording is much better :) nice change!  
  
I just was not 100% following the unlock rules, but also looking at the rust code (still learning rust) it seemed like this `Lock` would be useable with any `Type`, `Data` not matter so much (only reason it matters in the examples is because its used with the `Type` Script)  


> * As for changing the lock script name, further discussion may be required
All good if it remains, I understand the current name, but **Cheque** usually relates to "payments" of sorts and would be less "broad/generic" (strictly sUDT for example it makes sense)

duanyytop:
> I just was not 100% following the unlock rules, but also looking at the rust code (still learning rust) it seemed like this `Lock` would be useable with any `Type`, `Data` not matter so much (only reason it matters in the examples is because it's used with the `Type` Script requiring `Data` to be set)

Some `type scripts` depend on the specific `lock scripts`, like[CoTA](https://talk.nervos.org/t/rfc-cota-a-compact-token-aggregator-standard-for-extremely-low-cost-nfts-and-fts/6338) whose type script args must be equal to the lock hash when the CoTA cell is created and the type script args must not be changed anytime. 

So not all TYPE/DATA for cheque lock script, it depends on the type script, like CoTA type script.

rfcs/0039-cheque/0039-cheque.md
```diff
@@ -0,0 +1,334 @@
+---
+Number: "0039"
+Category: Standards Track
+Status: Proposal
+Author: Dylan Duan
```
janx, line 5: Please update metadata to latest format, thanks.

duanyytop: Updated. https://github.com/nervosnetwork/rfcs/pull/308/commits/e16494f9f3f3d7a938d6111be2ea90be1fe03e2b

janx (approved):

XuJiandong (changes requested):

jordanmack:
@duanyytop It would be beneficial to add requirements to the steps for the Reproducible Build. Without them, many errors will be received since it requires specific versions of a number of dependencies. My recommendation is to include the requirements for the most recent Ubuntu LTS version since that is the most common distro right now.

rfcs/0039-cheque/0039-cheque.md
```diff
@@ -0,0 +1,333 @@
...
+A cheque cell is the cell whose lock script is cheque script and its structure looks like following:
+
+```
+data:
+    amount: uint128
+type:
+    <simple_udt_type_script>
+lock:
+    code_hash: cheque lock script
+    args: <20 byte receiver secp256k1-blake2b-sighash-all lock hash> <20 byte sender secp256k1-blake2b-sighash-all lock hash>
```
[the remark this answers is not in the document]
duanyytop: Other RFC documents have many different descriptions: `20 byte`, `20 bytes`, and `20-byte`. I think we should have a specification first.

rfcs/0039-cheque/0039-cheque.md
```diff
@@ -0,0 +1,333 @@
...
+
+     - 1.b.ii. It loops through all output cells using the sender lock hash, if the sum of the output cells capacity is not equal to the sum of the cheque input cells capacity, the cheque lock returns with an error state
+
+   - 1.c. If the provided signature is valid with the sender secp256k1 public key hash, then:
+
+     - 1.c.i. It loops through all input cells using the current cheque lock script, if the `since` of any input cell is not the same as `0xA000000000000006` which means the tx failed verification unless it is 6 epochs later since the input cells get confirmed on-chain, the cheque lock returns with an error state
+
+2. If a signature is not provided in witness, then:
+
+   - 2.a. It loops through all input cells using the receiver and the sender lock hash, if no matching input cells are found, the cheque lock returns with an error state.
```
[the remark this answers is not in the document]
duanyytop: I used google translate and found that plural expressions are also ok.

duanyytop:
> It would be beneficial to add requirements to the steps for the Reproducible Build. Without them, many errors will be received since it requires specific versions of a number of dependencies. My recommendation is to include the requirements for the most recent Ubuntu LTS version since that is the most common distro right now

duanyytop:
> @duanyytop It would be beneficial to add requirements to the steps for the Reproducible Build. Without them, many errors will be received since it requires specific versions of a number of dependencies. My recommendation is to include the requirements for the most recent Ubuntu LTS version since that is the most common distro right now.

I think the [CI configure](https://github.com/nervosnetwork/ckb-cheque-script/blob/main/.github/workflows/build_and_test.yml) is enough and @jordanmack how do you feel?

jordanmack:
> I think the [CI configure](https://github.com/nervosnetwork/ckb-cheque-script/blob/main/.github/workflows/build_and_test.yml) is enough and @jordanmack how do you feel?

Yes, that is adequate. Just add a line that links to this and describes what it is.

duanyytop:
> > I think the [CI configure](https://github.com/nervosnetwork/ckb-cheque-script/blob/main/.github/workflows/build_and_test.yml) is enough and @jordanmack how do you feel?
> 
> Yes, that is adequate. Just add a line that links to this and describes what it is.

https://github.com/nervosnetwork/rfcs/pull/308

XuJiandong (dismissed):

janx (dismissed):

rfcs/0039-cheque/0039-cheque.md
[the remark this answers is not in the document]
jordanmack: In this context, I would recommend `20-byte`.

jordanmack:
@duanyytop I will be submitting a pull request with some additional suggestions for readability, but first I have some questions.

```
1.b.i. It loops through all input cells using the current cheque lock script, if any of the inputs' since is not zero, the cheque lock returns with an error state.
```

Unlock Rules 1.b.i will error if the receiver is valid and a since was used on one of the inputs. It seems like the transaction would still work if a since was included. What is the reasoning for disallowing the since here?

```
2.b. If the matching input cells with the receiver lock hash are found, it performs the same check as in 1.b.i and 1.b.ii.
2.b.i. It loops through all input cells using the receiver lock hash, if the matching inputs are not found, the cheque lock returns with an error state
```

Unlock Rules 2.b checks if any input cells are using the receiver lock and then performs validations from steps 1.b.i and 1.b.ii. Then in step 2.b.i it loops through these same input cells using the receiver lock, and checks for what? It sounds like it checks for the receiver lock cells, then loops through these cells to check if it has the receiver lock again. I do not understand this step.

```
2.b.ii It loops through all input cells using the receiver lock hash, if the first witness of the cheque cell group is empty(the witness is not WitnessArgs or the lock of WitnessArgs is empty)
```

Unlock Rules 2.b.ii appears to be incomplete. I do not understand this step.

duanyytop:
> Unlock Rules 1.b.i will error if the receiver is valid and a since was used on one of the inputs. It seems like the transaction would still work if a since was included. What is the reasoning for disallowing the since here?

For the receiver, when since is equal to 0, he/she can claim the asset without waiting for a period of time.

> Unlock Rules 2.b checks if any input cells are using the receiver lock and then performs validations from steps 1.b.i and 1.b.ii. Then in step 2.b.i it loops through these same input cells using the receiver lock, and checks for what? It sounds like it checks for the receiver lock cells, then loops through these cells to check if it has the receiver lock again. I do not understand this step.

This rule does seem redundant and can be deleted.

> Unlock Rules 2.b.ii appears to be incomplete. I do not understand this step.

Indeed, a paragraph should be added at the end: “the cheque lock returns with an error state“

jordanmack:
@duanyytop I'm rewriting part of the description for clarity. I want to make sure I understand the functionality and intent correctly.

The Cheque Lock can match against a `secp256k1-blake2b-sighash-all` signature, which means the default lock must be used when a signature is provided to unlock.

The Cheque Lock can also match against a lock hash from the inputs, in which case it can be any form of lock script since it is just a hash that is being checked.

Is the intent here to encourage the use of the default lock only, or should it be stated that smart contracts could also use the Cheque Lock when signatures are not provided?

duanyytop:
> @duanyytop I'm rewriting part of the description for clarity. I want to make sure I understand the functionality and intent correctly.
> 
> The Cheque Lock can match against a `secp256k1-blake2b-sighash-all` signature, which means the default lock must be used when a signature is provided to unlock.
> 
> The Cheque Lock can also match against a lock hash from the inputs, in which case it can be any form of lock script since it is just a hash that is being checked.
> 
> Is the intent here to encourage the use of the default lock only, or should it be stated that smart contracts could also use the Cheque Lock when signatures are not provided?

The Cheque Lock only supports `secp256k1-blake2b-sighash-all` signature , so the sender and receiver can only use `secp256k1-blake2b-sighash-all` lock script to build transactions.

jordanmack:
> The Cheque Lock only supports `secp256k1-blake2b-sighash-all` signature , so the sender and receiver can only use `secp256k1-blake2b-sighash-all` lock script to withdraw and claim the assets.

When no witness is provided, it attempts to match against the input lock hashes, as shown [here](https://github.com/duanyytop/ckb-cheque-script/blob/main/contracts/ckb-cheque-script/src/entry.rs#L45-L53). I do not know if we have an official name for this pattern. I've been calling it an "authorization piggyback". Since it is just matching the lock hash, it could be a lock of any type, including a lock script that is part of a smart contract.

1. Is my understanding of this correct?
2. Should I indicate that Cheque Lock only be used with `secp256k1-blake2b-sighash-all` to avoid complications, even if it could technically be used with different locks?

duanyytop:
> > The Cheque Lock only supports `secp256k1-blake2b-sighash-all` signature , so the sender and receiver can only use `secp256k1-blake2b-sighash-all` lock script to withdraw and claim the assets.
> 
> When no witness is provided, it attempts to match against the input lock hashes, as shown [here](https://github.com/duanyytop/ckb-cheque-script/blob/main/contracts/ckb-cheque-script/src/entry.rs#L45-L53). I do not know if we have an official name for this pattern. I've been calling it an "authorization piggyback". Since it is just matching the lock hash, it could be a lock of any type, including a lock script that is part of a smart contract.
> 
> 1. Is my understanding of this correct?
> 2. Should I indicate that Cheque Lock only be used with `secp256k1-blake2b-sighash-all` to avoid complications, even if it could technically be used with different locks?

```
lock:
    code_hash: cheque lock script
    args: <20 byte receiver secp256k1-blake2b-sighash-all lock hash> <20 byte sender secp256k1-blake2b-sighash-all lock hash>
```
The Cheque Lock script data structure determines that the lock hash you are talking about can only be `secp256k1-blake2b-sighash-all` lock hash. And the reason why only `secp256k1-blake2b-sighash-all` is supported is determined by the requirements at that time.

jordanmack:
@duanyytop A few more question in relation to Unlock Rule `2.b.ii`:

I see that within `claim::validate()` it [locates the witness for the receiver input cell](https://github.com/duanyytop/ckb-cheque-script/blob/main/contracts/ckb-cheque-script/src/claim.rs#L24-L29), but then within `helpers::check_witness_args()` it [does not validate the signature at all, and just ensures that any signature exists](https://github.com/duanyytop/ckb-cheque-script/blob/main/contracts/ckb-cheque-script/src/helper.rs#L41-L45).

1. What is the reason for checking that a signature exists, but not actually validating the signature?
2. Why check for a signature at all when doing an authorization piggyback since you are relying on the logic of the input's lock, and you cannot be certain that the lock is designed to require a signature to unlock?

duanyytop:
> 1. What is the reason for checking that a signature exists, but not actually validating the signature?

If the `WitnessArgs` is not empty for `secp256k1-blake2b-sighash-all`, the signature will be verified by the receiver lock script and Cheque Lock does not need to do repetitive things.

> 2. Why check for a signature at all when doing an authorization piggyback since you are relying on the logic of the input's lock, and you cannot be certain that the lock is designed to require a signature to unlock?

I'm sorry, I don't quite understand what you mean

jordanmack:
> If the `WitnessArgs` is not empty for `secp256k1-blake2b-sighash-all`, the signature will be verified by the receiver lock script and Cheque Lock does not need to do repetitive things.

Yes, this is the crux of my question. The Cheque Lock does not need to do a repetitive signature validation that is already done by the receiver lock script. So why does it check that `WitnessArgs` is not empty? What is ensured by doing this extra step?

duanyytop:
> > If the `WitnessArgs` is not empty for `secp256k1-blake2b-sighash-all`, the signature will be verified by the receiver lock script and Cheque Lock does not need to do repetitive things.
> 
> Yes, this is the crux of my question. The Cheque Lock does not need to do a repetitive signature validation that is already done by the receiver lock script. So why does it check that `WitnessArgs` is not empty? What is ensured by doing this extra step?



> > If the `WitnessArgs` is not empty for `secp256k1-blake2b-sighash-all`, the signature will be verified by the receiver lock script and Cheque Lock does not need to do repetitive things.
> 
> Yes, this is the crux of my question. The Cheque Lock does not need to do a repetitive signature validation that is already done by the receiver lock script. So why does it check that `WitnessArgs` is not empty? What is ensured by doing this extra step?

Because there is no hard-coded `secp256k1-blake2b-sighash-all` lock information in the contract, that is to say, we cannot guarantee that the user must use the `secp256k1-blake2b-sighash-all` lock. Checking whether the `WitnessArgs` is empty can avoid the risk of evil to a certain extent (although it cannot be fundamentally eliminated)

jordanmack:
I have created a new pull request to @duanyytop's branch with my suggested edits.

Preview: https://github.com/jordanmack/duanyytop-rfcs/blob/add-cheque-lock/rfcs/0039-cheque/0039-cheque.md
Pull Request: https://github.com/duanyytop/rfcs/pull/4
Commit Details: https://github.com/jordanmack/duanyytop-rfcs/commit/2b6b028ddf8a7df10288760a977e5aa9f1f6a0e4

duanyytop:
@jordanmack any questions?

jordanmack (approved):

jordanmack:
> @jordanmack any questions?

No more questions. Everything should be included in the PR.

janx (approved):

XuJiandong (approved):

PR #343, opened 2022-05-25
RFC: Omnilock

XuJiandong:


rfcs/0000-omnilock/0000-omnilock.md
```diff
@@ -0,0 +1,558 @@
...
+        Lock:
+            code_hash: Omnilock
+            args: <flag: 0xFE> <preimage hash> <Omnilock flags: 0>
+    <...>
+Outputs:
+    <vec> Any cell
+Witnesses:
+    WitnessArgs structure:
+      Lock:
+        signature: <valid secp256k1 signature for pubkey hash 1>
```
homura, line 426:
```suggestion
        signature: <valid delegate signature for pubkey hash 1>
```

I found that in the current omnilock implementation, there is [no restriction](https://github.com/nervosnetwork/ckb-production-scripts/blob/master/c/ckb_identity.h#L343-L344) that the cryptographic algorithm must be secp256k1 elliptic curve, or even [other cryptographic algorithms](https://github.com/nervosnetwork/ckb-production-scripts/blob/master/tests/omni_lock_rust/tests/misc.rs#L633-L635)

rfcs/0000-omnilock/0000-omnilock.md
```diff
@@ -0,0 +1,558 @@
...
+        Lock:
+            code_hash: Omnilock
+            args: <flag: 0xFD> <preimage hash> <Omnilock flags: 0>
+    <...>
+Outputs:
+    <vec> Any cell
+Witnesses:
+    WitnessArgs structure:
+      Lock:
+        signature: <valid secp256k1 signature for pubkey hash 1>
```
homura, line 449:
```suggestion
        signature: <valid delegate signature for pubkey hash 1>
```

rfcs/0000-omnilock/0000-omnilock.md
XuJiandong: This is an example so that it's described as detailed as possible.

rfcs/0042-omnilock/0042-omnilock.md
```diff
@@ -0,0 +1,558 @@
...
+
+When "administrator mode" is enabled, `<32 byte AdminList cell Type ID>` must be present. The AdminList cell contains the
+type script hash used by a special cell with the same format as [RCE
+Cell](https://talk.nervos.org/t/rfc-regulation-compliance-extension/5338). The RCE cell follows a set of rules and contains
+whitelists and blacklists. These lists can be used in the [SMT proofs
+scenarios](https://github.com/nervosnetwork/sparse-merkle-tree).
+
+The RCE cells are organized in tree structure illustrated in the following diagram:
+
+![RCE Cells](./rce_cells.png)
```
doitian, line 93: Please share the diagram source file in comment.

XuJiandong: https://drive.google.com/file/d/1mKgqIl2IK-n9JV0kktT2zZDvDaRsUS7D/view?usp=sharing
Open it with draw.io

rfcs/0000-omnilock/0000-omnilock.md
homura: as a newcomer, I may mistakenly think that Omnilock can only support secp256k1, since secp256k1 is very common in CKB, maybe we can replace this example with RSA to explain the meaning of **delegate signature**?

XuJiandong: As an example, it should be explained as detailed as possible. This is not only applied to this RFC but also other RFCs.
To understand RFC fully, newcomers need to read chapters above.

rfcs/0042-omnilock/0042-omnilock.md
```diff
@@ -0,0 +1,558 @@
...
+Number: "0042"
+Category: Standards Track
+Status: Proposal
+Author: Xu Jiandong <lynndon@gmail.com>
+Created: 2022-05-19
+---
+
+# Omnilock
+
+Omnilock is a new lock script designed for interoperability. It supports various transaction verifications used in
```
janx, line 11:
```suggestion
Omnilock is a lock script designed for interoperability. It comes with built-in support for verification of transaction signing methods used in
```

rfcs/0042-omnilock/0042-omnilock.md
```diff
@@ -0,0 +1,558 @@
...
+popular blockchains, such as Bitcoin, Ethereum, EOS, and Dogecoin. Omnilock is also extensible, so more verification
+algorithms can be added in future. This feature makes Omnilock a powerful module in Nervos interoperability 2.0.
```
janx, line 13:
```suggestion
algorithms can be added in future.
```

rfcs/0042-omnilock/0042-omnilock.md
```diff
@@ -0,0 +1,558 @@
...
+
+Another feature of Omnilock for practitioners is the regulation compliance module (Consider it as interoperability with
```
janx, line 15:
```suggestion
Another feature of Omnilock for practitioners is the regulation compliance module which brings interoperability with
```

rfcs/0042-omnilock/0042-omnilock.md
```diff
@@ -0,0 +1,558 @@
...
+the traditional world). If enabled, the specified administrator can revoke tokens held by users under circumstances
+which the administrator deems proper. This part has evolved from the [Regulation Compliance Extension
+(RCE)](https://talk.nervos.org/t/rfc-regulation-compliance-extension/5338) proposal for
+[xUDT](https://talk.nervos.org/t/rfc-extensible-udt/5337). This feature provides an option that sits at the other side
+of the asset lock spectrum and lays the foundation of registered assets like Apple stock on CKB. When used together,
+Omnilock and RCE provide an [ERC-1404](https://erc1404.org/) equivalence.
+
+## Authentication
+
+Omnilock introduces a new concept, authentication ( auth ) to CKB lock scripts: an auth is a 21-byte data structure
```
janx, line 25:
```suggestion
An authentication (auth) is a 21-byte data structure
```

janx (changes requested):
I walked through the open section and removed decorative words to make the RFC more concise. Some paragraphs may need reformat.

rfcs/0042-omnilock/0042-omnilock.md
```diff
@@ -0,0 +1,558 @@
...
+
+* When the auth flag is 0xFC, we will check against the current transaction, and there must be an input cell, whose lock
+  script matches the auth content when hashed via blake160.
+
+
+When `signature` is present, the signature can be used to unlock the cell in anyone-can-pay mode.
+
+When `preimage` is present, the auth flag can be 0xFD or 0xFE.
+
+* When auth flag is 0xFD, the `exec` method is used. The preimage's memory layout will be as follows:
```
quake, line 202: suggest to add a prefix to preimage layout, it can be used as a tag or extension bit:

| Flag | Preimage Layout | Verification |
| ------------- | ------------- | ------------- |
| 0x01  | exec code hash(32 bytes) exec hash type (1 byte)...  | lock.args[0] == 0xFD && lock.args[1..21] == blake2b(preimage)|
| 0x02  | dynamic library code hash (32 bytes) ...| lock.args[0] == 0xFE && lock.args[1..21] = blake2b(preimage)|

rfcs/0042-omnilock/0042-omnilock.md
```diff
@@ -0,0 +1,558 @@
...
+bounds (8 bytes)
+pubkey blake160 hash (20 bytes)
+```
+
+The `place` and `bounds` are passed in directly to [exec](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0034-vm-syscalls-2/0034-vm-syscalls-2.md#exec):
+```C
+int ckb_exec(size_t index, size_t source, size_t place, size_t bounds, int argc, char* argv[]);
+```
+The `index` is located by `exec code hash` and `exec hash type` from `cell_deps`([See usage](https://github.com/nervosnetwork/ckb-c-stdlib/blob/20578dfb092b3b3761df755395e20ec142a83d6e/ckb_syscalls.h#L368)].
+Finally, message, signature, pubkey blake160 hash are encoded into hex strings. Then these hex strings are passed in as `argv`.
```
quake, line 216: What does the `message` here mean specifically? Can you add more descriptions?

rfcs/0042-omnilock/0042-omnilock.md
```diff
@@ -0,0 +1,558 @@
...
+```
+
+It loads the dynamic linking libraries via `code hash` and `hash type`, and gets the entry function named 
+`validate_signature`. The entry function is expected to have following C API:
+```C
+int validate_signature(void *prefilled_data, const uint8_t *signature_buffer,
+    size_t signature_size, const uint8_t *message_buffer, size_t message_size,
+    uint8_t *pubkey_hash, size_t *pubkey_hash_len);
+```
+Then the entry function is called to validate the message and signature. The `pubkey_hash` returned from the entry
```
quake, line 232: same question, what does the message here mean specifically?

rfcs/0042-omnilock/0042-omnilock.md
```diff
@@ -0,0 +1,558 @@
...
+exec code hash (32 bytes)
+exec hash type (1 byte)
+place (1 byte)
+bounds (8 bytes)
+pubkey blake160 hash (20 bytes)
+```
+
+The `place` and `bounds` are passed in directly to [exec](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0034-vm-syscalls-2/0034-vm-syscalls-2.md#exec):
+```C
+int ckb_exec(size_t index, size_t source, size_t place, size_t bounds, int argc, char* argv[]);
```
quake, line 213: Since the source can only be cell deps, why do we need to set the place parameter? It is only legal if it is 0, suggest to remove the `place (1 byte)` from the preimage

quake (changes requested):

rfcs/0042-omnilock/0042-omnilock.md
XuJiandong: This change requires the implementation to be updated. But the implementation has been deployed.
If we have further version of omnilock, I think it's OK to accept this change.

rfcs/0042-omnilock/0042-omnilock.md
XuJiandong: The description of the `message` is added in the section `Omnilock Witness`.

quake (approved):

janx (dismissed):

zhangsoledad (dismissed):

janx (approved):

zhangsoledad (approved):

quake (approved):

doitian (approved):

PR #349, opened 2022-06-16
RFC(0043): CKB softfork activation

### LINKED ISSUE #3630: todo list of release 0.105.0
- [x] create a PR https://github.com/nervosnetwork/ckb/pull/3631 to merge branch light-client-rebase to develop and code review
- [ ] review 3 rfcs
    - [ ] https://github.com/nervosnetwork/rfcs/pull/349
    - [ ] https://github.com/nervosnetwork/rfcs/pull/370
    - [ ] https://github.com/nervosnetwork/rfcs/pull/375

zhangsoledad:
[Preview Link](https://github.com/zhangsoledad/rfcs/blob/soft-fork-activation/rfcs/0043-ckb-softfork-activation/0043-ckb-softfork-activation.md)
This is a proposal based on BIP9 modified by ckb characteristics and also incorporates the "speedy trial" activation specification.


rfcs/0000-ckb-softfork-activation/0000-ckb-softfork-activation.md
```diff
@@ -0,0 +1,132 @@
...
+
+1. DEFINED is the first state that each softfork starts. The genesis epoch is by definition in this state for each deployment.
+2. STARTED for epochs past the `start_epoch`.
+3. LOCKED_IN for one epoch after the first epoch period with STARTED epochs of which at least `threshold` has the associated bit set in `version`.
+4. ACTIVE for all epochs after the LOCKED_IN epoch.
+5. FAILED for one epoch period past the `timeout_epoch`, if LOCKED_IN was not reached.
+
+### Bit flags
+
+The `version` block header field is to be interpreted as a 32-bit little-endian integer (as present), and bits are selected within this integer as values (1 << N) where N is the bit number.
```
janx, line 51: > ... and bits are selected within this integer as values (1 << N) where N is the bit number.

Not sure what this sentence tries to express here.

rfcs/0000-ckb-softfork-activation/0000-ckb-softfork-activation.md
```diff
@@ -0,0 +1,132 @@
...
+
+The genesis epoch has a state DEFINED for each deployment, by definition.
+
+```rust
+if epoch.is_genesis() {
+    return ThresholdState::DEFINED;
+}
+```
+
+We remain in the initial state until we reach the `start_epoch`.
```
janx, line 74:
```suggestion
The deployment remains in the initial state until we reach the `start_epoch`.
```

rfcs/0000-ckb-softfork-activation/0000-ckb-softfork-activation.md
```diff
@@ -0,0 +1,132 @@
...
+Miners should continue setting the bit in the LOCKED_IN phase, so uptake is visible, though this does not affect consensus rules.
+
+### New consensus rules
+The new consensus rules for each softfork are enforced for each block with an ACTIVE state.
+
+### State transitions
+
+![State transitions](images/state-transitions.png)
+
+The genesis epoch has a state DEFINED for each deployment, by definition.
```
janx, line 66: It seems "genesis epoch" means the beginning epoch of a softfork proposal in this RFC? If so I think it's a bad name 'cause it's easy to mix it with the epoch 0 of CKB chain.

rfcs/0000-ckb-softfork-activation/0000-ckb-softfork-activation.md
```diff
@@ -0,0 +1,132 @@
...
+            }
+        }
+        let threshold_number = threshold_number(epoch_length, threshold);
+        if count >= threshold_number {
+            next_state = ThresholdState::LOCKED_IN;
+        } else if epoch_ext.number() >= timeout {
+            next_state = ThresholdState::FAILED;
+        }
+    }
+```
```
janx, line 105: Is this check for `LOCKED_IN` run at the beginning of each epoch, or at each block?

rfcs/0000-ckb-softfork-activation/0000-ckb-softfork-activation.md
```diff
@@ -0,0 +1,132 @@
...
+
+After at least one epoch of LOCKED_IN, we automatically transition to ACTIVE if the `minimum_activation_epoch` is reached. Otherwise, LOCKED_IN continues.
```
janx, line 107:
```suggestion
After at least one epoch of LOCKED_IN, the deployment automatically transits to ACTIVE if the `minimum_activation_epoch` is reached. Otherwise, LOCKED_IN continues.
```

rfcs/0000-ckb-softfork-activation/0000-ckb-softfork-activation.md
```diff
@@ -0,0 +1,132 @@
...
+Furthermore, ACTIVE and FAILED are terminal states in which a deployment stays once reached.
+
+```rust
+ThresholdState::FAILED | ThresholdState::ACTIVE => {
+    // Nothing happens, these are terminal states.
+}
+```
+
+### Warning mechanism
+To support upgrade warnings, an extra "unknown upgrade" is tracked, using the "implicit bit" mask = (version & ~expected_version) != 0. Mask will be non-zero whenever an unexpected bit is set in `version`. Whenever LOCKED_IN for the unknown upgrade is detected, the software should warn loudly about the upcoming softfork. It should warn even more loudly after the next epoch (when the unknown upgrade is in the ACTIVE state).
```
janx, line 126: Does "the software" mean CKB node or wallet or something else? What does "warn loudly" mean?

> It should warn even more loudly after the next epoch (when the unknown upgrade is in the ACTIVE state).

Why the unknown upgrade is in ACTIVE in the "next epoch" of LOCKED_IN?

janx: From my understanding, such an "unknown upgrade" can only triggered by miners setting a bit in `version` when that bit has no corresponding deployment info encoded in the CKB node. If so it can never be ACTIVE because it has no `minimum_activation_epoch' defined.

rfcs/0000-ckb-softfork-activation/0000-ckb-softfork-activation.md
```diff
@@ -0,0 +1,132 @@
...
+### Parameters
+
+Each softfork deployment is specified by the following per-chain parameters (further elaborated below):
+
+1. The `name` specifies a very brief description of the softfork, reasonable for use as an identifier.
+2. The `bit` determines which bit in the `version` field of the block is to be used to signal the softfork lock-in and activation. It is chosen from the set {0,1,2,...,28}.
+3. The `start_epoch` specifies the first epoch in which the bit gains meaning.
+4. The `timeout_epoch` specifies an epoch at which the miner signaling ends. Once this epoch has been reached, if the softfork has not yet locked_in (excluding this epoch block's bit state), the deployment is considered failed on all descendants of the block.
+5. The `threshold` specifies the minimum ratio of block per epoch, which indicate the locked_in of the softfork during the epoch.
+6. The `minimum_activation_epoch` specifies the epoch at which the softfork is allowed to become active.
```
janx, line 26: I suggest rename `minimum_activation_epoch` to `activation_epoch`. With a `minimum_` prefix, it implies "the softfork can be activated at any epoch after this epoch". But the wanted behavior is to activate the softfork immediately at this activation epoch.

janx (changes requested):
An epoch in CKB (4 hours) is far shorter than an epoch in BItcoin (2 weeks). The tally period of softfork signaling should be long enough so it's resistant to governance attack. Instead of 1 epoch (4 hours), I think CKB softfork activation tally period should be at least as long as Bitcoin's (2 weeks ~= 90 CKB epochs).

rfcs/0000-ckb-softfork-activation/0000-ckb-softfork-activation.md
zhangsoledad: No, it does means the epoch 0 of CKB chain.
This proposal allowing multiple deploy in parallel, so multiple parallel deployments will have different states, but we specify that each deployment will start from DEFINED state, with genesis as the starting point, until reach the specified start_epoch, then the state will change to started.

rfcs/0000-ckb-softfork-activation/0000-ckb-softfork-activation.md
```diff
@@ -0,0 +1,132 @@
...
+4. `timeout_epoch` should be set to an epoch when it is considered reasonable to expect the entire economy to have upgraded by, probably a month, 180 epochs after `start_epoch`.
+5. `threshold` should be 90% or 75% for testnet.
+6. `minimum_activation_epoch` should be set to several epochs after `timeout_epoch` if the `start_epoch` is to be very soon after software with parameters is expected to be released.
+Where the locked_in threshold is reached, softforks are guaranteed to activate eventually but not until `minimum_activation_epoch` after signal tracking starts, allowing users, developers, and organizations to prepare software, announcements, and celebrations for that event.
+
+### States
+
+With each epoch and softfork, we associate a deployment state. The possible states are:
+
+1. DEFINED is the first state that each softfork starts. The genesis epoch is by definition in this state for each deployment.
```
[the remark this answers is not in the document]
zhangsoledad: Maybe 0 epoch more clear.

rfcs/0000-ckb-softfork-activation/0000-ckb-softfork-activation.md
zhangsoledad: The state of the blocks in the same epoch is the same, and the state of the epoch depends on the previous epoch, i.e. we only count the previous epoch. So this check will only be done once per epoch and we cache cache the results.

zhangsoledad:
> An epoch in CKB (4 hours) is far shorter than an epoch in BItcoin (2 weeks). The tally period of softfork signaling should be long enough so it's resistant to governance attack. Instead of 1 epoch (4 hours), I think CKB softfork activation tally period should be at least as long as Bitcoin's (2 weeks ~= 90 CKB epochs).

I'm not so sure it's appropriate to measure the replacement in realistic time, CKB is much faster to mine, 2 weeks ~= 151,200 blocks.
A big problem with BIP9 is that without the cooperation of miners, the process would be relatively long and a year would be wasted if it failed. So `Speedy Trial` proposes to "Start soon, Stop soon", there may have room for compromise here, for period and timeout.

rfcs/0000-ckb-softfork-activation/0000-ckb-softfork-activation.md
janx: Just found that `minimum_activation_epoch` is actually accurate after reading BIP8 and BIP9 again.

janx:
> > An epoch in CKB (4 hours) is far shorter than an epoch in BItcoin (2 weeks). The tally period of softfork signaling should be long enough so it's resistant to governance attack. Instead of 1 epoch (4 hours), I think CKB softfork activation tally period should be at least as long as Bitcoin's (2 weeks ~= 90 CKB epochs).
> 
> I'm not so sure it's appropriate to measure the replacement in realistic time, CKB is much faster to mine, 2 weeks ~= 151,200 blocks. A big problem with BIP9 is that without the cooperation of miners, the process would be relatively long and a year would be wasted if it failed. So `Speedy Trial` proposes to "Start soon, Stop soon", there may have room for compromise here, for period and timeout.

Why is `lockinontimeout` not included in this proposal?

rfcs/0000-ckb-softfork-activation/0000-ckb-softfork-activation.md
zhangsoledad: There is a similar discussion here, https://github.com/bitcoin/bitcoin/pull/21377/files#r598896420

zhangsoledad:
> `lockinontimeout`

`lockinontimeout` is the logic of BIP8, which is used to start a flag day.

>  * MUST_SIGNAL for one retarget period prior to the timeout, if LOCKED_IN was not reached and `lockinontimeout` is true.
>  * During the MUST_SIGNAL phase, if (2016 - threshold) blocks in the retarget period have already failed to signal, any further blocks that fail to signal are invalid.

Optimistically, we hope that BIP8 will not be used, 
BIP8 forced signaling (or automatic activation) in specific, is that it can be risky, this scheme could split the network between upgraded and non-upgraded nodes.
> This risk is probably best countered by offering enough time to upgrade. Unfortunately, not everyone agrees on how much time is enough.

rfcs/0000-ckb-softfork-activation/0000-ckb-softfork-activation.md
zhangsoledad: Remove this section for now.

zhangsoledad:
The current signaling period is tentatively set at 10 epochs.

README.md
```diff
@@ -84,6 +84,8 @@ The maintainers of RFCs and the community will review the PR, and you should upd
 | [27](rfcs/0027-block-structure) | [CKB Block Structure](rfcs/0027-block-structure/0027-block-structure.md) | Ian Yang | Informational | Draft
 | [37](rfcs/0037-ckb2021) | [CKB Consensus Change (Edition CKB2021)](rfcs/0037-ckb2021/0037-ckb2021.md) | Ian Yang | Informational | Draft
 | [42](rfcs/0042-omnilock) | [Omnilock](rfcs/0042-omnilock/0042-omnilock.md) | Xu Jiandong | Standards Track | Proposal
+| [43](rfcs/0043-ckb-softfork-activation) | [CKB Softfork Activation](rfcs/0043-ckb-softfork-activation/0043-ckb-softfork-activation) | Dingwei Zhang | Standards Track | Draft
```
yangby-cryptape, line 87:
```suggestion
| [43](rfcs/0043-ckb-softfork-activation) | [CKB Softfork Activation](rfcs/0043-ckb-softfork-activation/0043-ckb-softfork-activation.md) | Dingwei Zhang | Standards Track | Draft
```

rfcs/0043-ckb-softfork-activation/0043-ckb-softfork-activation.md
```diff
@@ -0,0 +1,134 @@
...
+    // Nothing happens, these are terminal states.
+}
+```
+
+
+## Reference
+
+* https://github.com/bitcoin/bips/blob/master/bip-0009.mediawiki
+* https://github.com/bitcoin/bips/blob/master/bip-0008.mediawiki
+* https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2021-March/018583.html
```
doitian, line 134: Suggest changes:

- Cite the title and authors
- User an orderd list

```suggestion
1. Wuille, P., Todd, P., Maxwell, G., & Russell, R. (2015). BIP9: Version bits with timeout and delay. Bitcoin BIPs. https://github.com/bitcoin/bips/blob/master/bip-0009.mediawiki
2. Fry, S., & Dashjr, L. (2017). BIP8: Version bits with lock-in by height. Bitcoin BIPs. https://github.com/bitcoin/bips/blob/master/bip-0008.mediawiki
3. Harding, D. A. (2021). Taproot activation proposal “Speedy Trial.” Bitcoin-Dev Mailing List. https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2021-March/018583.html
```

rfcs/0043-ckb-softfork-activation/images/.keep
```diff
@@ -0,0 +1 @@
+
```
doitian, line 1: The file is not necessary.

rfcs/0043-ckb-softfork-activation/0043-ckb-softfork-activation.md
```diff
@@ -0,0 +1,134 @@
...
+```rust
+match state {
+    ThresholdState::STARTED => {
+        let mut count = 0;
+        for block in (0..period_blocks) {
+            if (block.version() & 0xE0000000 == 0x20000000 && (block.version() >> bit) & 1 == 1) {
+                ++count;
+            }
+        }
+        let threshold_number = threshold_number(epoch_length, threshold);
```
doitian, line 102: Is `epoch_length` wrong here? I think `period_blocks` should be used here.

doitian (approved):

janx (approved):

PR #356, opened 2022-07-26
Remove reference to maturity constraint

phroi:
It has been a while that the [RFC 0036 Remove Header Deps Immature Rule](../0036-remove-header-deps-immature-rule/0036-remove-header-deps-immature-rule.md) has been in effect, it should be safe to remove the old reference to the cell maturity constraint.


quake (approved):

zhangsoledad (approved):

doitian (approved):

janx:
@phroi Thanks for the PR :beers:

PR #418, opened 2023-04-22
VM Syscalls 3

zhangsoledad:


rfcs/0147-vm-syscalls-3/0147-vm-syscalls-3.md
```diff
@@ -0,0 +1,158 @@
...
+This syscall might return the following results:
+
+- 0: Success.
+- 1-3: Reserved. These values are already assigned to other syscalls.
+- 4: Elf format error
+- 5: Exceeded max content length.
+- 6: Wrong memory limit
+- 7: Exceeded max peak memory
+
+Note that now we have a new limit called *Peak Memory Usage*. The maximum memory usage of the parent script and its descendant ********children cannot exceed this value. Currently this limit is set at 32M.
```
eval-exec, line 69: What's the meaning of `********children`.

rfcs/0147-vm-syscalls-3/0147-vm-syscalls-3.md
```diff
@@ -0,0 +1,188 @@
...
+Created: 2023-04-17
+---
+
+# VM Syscalls 3
+
+## Abstract
+
+This document describes the addition of the syscalls during the CKB2023. This update significantly enhances the flexibility of CKB Script.
+
+The following three syscalls are added:
```
mohanson, line 15: three => four

rfcs/0147-vm-syscalls-3/0147-vm-syscalls-3.md
```diff
@@ -0,0 +1,188 @@
...
+- [Set Content]
+- [Load Extension]
+
+### Spawn
+[Spawn]: #spawn
+
+The syscall Spawn is the core part of this update. The *Spawn* and the latter two syscalls: *Get Memory Limit* and *Set Content* together, implement a way to call another CKB Script in a CKB Script. Unlike the *Exec*[1](../0034-vm-syscalls-2/0034-vm-syscalls-2.md) syscall, *Spawn* saves the execution context of the current script, like [posix_spawn](https://man7.org/linux/man-pages/man3/posix_spawn.3.html), the parent script blocks until the child script ends.
+
+```c
+int ckb_spawn(uint64_t memory_limit, size_t index, size_t source,
```
mohanson, line 28: Note that we have updated the signature of spawn.  Please update it from internal docs.

zhangsoledad: @mohanson updated

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -0,0 +1,202 @@
...
+### Get Memory Limit
+[Get Memory Limit]: #get-memory-limit
+
+Get the maximum available memory for the current script.
+
+```c
+int ckb_get_memory_limit();
+```
+
+For the prime script, it will always return 8(4M). For the child script, it depends on the parameters set by *Spawn*.
```
XuJiandong, line 90:
```suggestion
For the script(prime script) directly invoked by CKB, it will always return 8(4M). For the child script invoked by prime script or other child script, it depends on the parameters set by *Spawn*.
```

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -0,0 +1,202 @@
...
+int ckb_set_content(uint8_t* content, uint64_t* length);
+```
+
+- Length up to 256K.
+- If the written length is greater than the limit given by *Spawn*, the final written length is the minimum of the two.
+- This function is optional. Not every child script needs to call this.
+
+### Spawn example
+
+Suppose we write a dependency library, the function of this library is very simple: receive parameters, then concatenates the parameters together and return to the caller.
```
XuJiandong, line 107:
```suggestion
Consider the creation of a dependency library with a straightforward function that receives parameters, concatenates them, and subsequently returns the resulting string to the caller.
```

XuJiandong (changes requested):

yangby-cryptape (commented):
If I understood correctly, all `M` in this RFC means $2^{20}$; strictly speaking, they should be `Mi`.

According to "ISO/IEC 80000":
- `M` is "mega", it's a [metric prefix], it means $10^{6}$.
- To represent $2^{20}$, [IEC prefix] `Mi` should be used, it's short for "mebi".

[metric prefix]: https://en.wikipedia.org/wiki/Metric_prefix
[IEC prefix]: https://en.wikipedia.org/wiki/Binary_prefix#IEC_prefixes

mohanson:
Please add the introduction of `Current Memory` syscall

### Current Memory

Get the Peak Memory Usage.

```c
int ckb_current_memory();
```

The system call returns an integer. Multiply it by 0.5 to get the actual memory usage (unit: megabytes)

zhangsoledad:
> Please add the introduction of `Current Memory` syscall
> 
> ### Current Memory
> Get the Current Memory Usage. The result is the sum of the memory usage of the parent script and the child script.
> 
> ```c
> int ckb_current_memory();
> ```
> 
> The system call returns an integer. Multiply it by 0.5 to get the actual memory usage (unit: megabytes)



> Please add the introduction of `Current Memory` syscall
> 
> ### Current Memory
> Get the Current Memory Usage. The result is the sum of the memory usage of the parent script and the child script.
> 
> ```c
> int ckb_current_memory();
> ```
> 
> The system call returns an integer. Multiply it by 0.5 to get the actual memory usage (unit: megabytes)

Already added, please review

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -0,0 +1,214 @@
...
+### Get Memory Limit
+[Get Memory Limit]: #get-memory-limit
+
+Get the maximum available memory for the current script.
+
+```c
+int ckb_get_memory_limit();
+```
+
+For the script(prime script) directly invoked by CKB, it will always return 8(4M). For the child script invoked by prime script or other child script, it depends on the parameters set by *Spawn*.
```
doitian, line 91: Lack of the explaination of the return result unit. There's only one example that 8 means 4M.

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -0,0 +1,214 @@
...
+
+The system call returns an integer. Multiply it by 0.5 to get the actual memory usage (unit: megabytes)
+
+### Set Content
+[Set Content]: #set-content
+
+The child script can return bytes data to the parent script through `Set Content`.
+
+```c
+int ckb_set_content(uint8_t* content, uint64_t* length);
```
doitian, line 110: What's the return result of this syscall? What happens when it is called in the root script?

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -0,0 +1,214 @@
...
+    - 1 (0.5 M)
+    - 2 (1 M)
+    - 3 (1.5 M)
+    - 4 (2 M)
+    - 5 (2.5 M)
+    - 6 (3 M)
+    - 7 (3.5 M)
+    - 8 (4 M)
+- `exit_code`: an int8 pointer denoting where we save the exit code of a child script.
+- `content`: a pointer to a buffer in VM memory space denoting where we would load the sub-script data. The child script will write data in this buffer via `set_content`.
```
doitian, line 63: What happens when passing the pointer 0 as content?

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -0,0 +1,214 @@
...
+- `content_length`: a pointer to a 64-bit unsigned integer in VM memory space. When calling the syscall, this memory location should store the length of the buffer specified by `content`. When returning from the syscall, CKB VM would fill in `content_length` with the actual length of the buffer. `content_length` up to 256K.
```
doitian, line 64: What happens when passing the pointer 0 as `content_length`?

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
doitian: Leave a reference to the document on `memory_limit`

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -0,0 +1,214 @@
...
+### Current Memory
+[Current Memory]: #current-memory
+
+Get the Current Memory Usage. The result is the sum of the memory usage of the parent script and the child script.
+
+```c
+int ckb_current_memory();
+```
+
+The system call returns an integer. Multiply it by 0.5 to get the actual memory usage (unit: megabytes)
```
doitian, line 102: Leave a reference to the document on `memory_limit`

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
mohanson: The specific algorithm of memory_limit is introduced above.

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
mohanson: If the subscript does not return data, then using 0 or NULL is correct; if the subscript returns data and 0 or NULL is passed in at this time, unexpected memory errors may occur.

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
mohanson: Depends on whether the subscript has called `set_content`. In one case, nothing will happen. In another case, a number may be written to 0.

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
mohanson: It returns 0(success) or others. Nothing happens when calling set_content in the root script.

XuJiandong (approved):

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -0,0 +1,217 @@
...
+- [Spawn]
+- [Get Memory Limit]
+- [Current Memory]
+- [Set Content]
+- [Load Block Extension]
+
+### Spawn
+[Spawn]: #spawn
+
+The syscall Spawn is the core part of this update. The *Spawn* and the latter two syscalls: *Get Memory Limit* and *Set Content* together, implement a way to call another CKB Script in a CKB Script. Unlike the *Exec*[1](../0034-vm-syscalls-2/0034-vm-syscalls-2.md) syscall, *Spawn* saves the execution context of the current script, like [posix_spawn](https://man7.org/linux/man-pages/man3/posix_spawn.3.html), the parent script blocks until the child script ends.
```
doitian, line 26: I think it's better to emphasize that the parent script continues instead of that it blocks.
```suggestion
The syscall Spawn is the core part of this update. The *Spawn* and the latter two syscalls: *Get Memory Limit* and *Set Content* together, implement a way to call another CKB Script in a CKB Script. Unlike the *Exec*[1](../0034-vm-syscalls-2/0034-vm-syscalls-2.md) syscall, *Spawn* saves the execution context of the current script, like [posix_spawn](https://man7.org/linux/man-pages/man3/posix_spawn.3.html), the parent script continues when the child script ends.
```

mohanson (approved):

XuJiandong (approved):

doitian (approved):

PR #428, opened 2024-01-09
[0052] Extensible UDT

XuJiandong:
Extensible UDT(xUDT) is an extension of Simple UDT for defining more behaviors a UDT might need. While simple UDT provides a minimal core for issuing UDTs on Nervos CKB, extensible UDT builds on top of simple UDT for more potential needs, such as regulations.

### DIFF
--- a/rfcs/0052-extensible-udt/0052-extensible-udt.md
+++ b/rfcs/0052-extensible-udt/0052-extensible-udt.md
@@ -211,15 +211,15 @@ Outputs:
     <...>
 Witnesses:
     WitnessArgs structure:
-    Lock: <user defined>
-    Input Type: <XudtWitnessInput>
-        owner_script: <None>
-        owner_signature: <None>				
-        raw_extension_data: <None>
-        extension_data: 
-            <vec> BytesVec
-                <data> 
-            <...>
+        Lock: <user defined>
+        Input Type: <XudtWitnessInput>
+            owner_script: <None>
+            owner_signature: <None>				
+            raw_extension_data: <None>
+            extension_data: 
+                <vec> BytesVec
+                    <data> 
+                <...>
 ```
 
 The witness of the same index as the first input xUDT cell is located by xUDT script. It is parsed first as WitnessArgs structure, the `input_type` or `output_type` field of `WitnessArgs`, is thus treated as `XudtWitnessInput` structure. The `extension_data`structure inside must also be of the same length as `xUDT args`. An extension script might also require transaction-specific data so as to validate. Witness here provides a place for this data needs.
--- a/rfcs/0052-extensible-udt/0052-extensible-udt.md
+++ b/rfcs/0052-extensible-udt/0052-extensible-udt.md
@@ -314,8 +314,7 @@ Witnesses:
                 <...>
 ```
 
-The example above shows a scenario of owner mode without consuming cell. This
-transaction can be valid even without any input cell. The `<owner lock script
+The example above shows a scenario of owner mode without consuming the owner's cell. The `<owner lock script
 hash 1>` is the same as blake160 hash of `<owner script 1>`. We can implement an
 extension script as `<owner script 1>` which has same functionality to
 [secp256k1/blake160 lock
--- a/rfcs/0052-extensible-udt/0052-extensible-udt.md
+++ b/rfcs/0052-extensible-udt/0052-extensible-udt.md
@@ -351,8 +351,8 @@ chosen to use a zero lock when deploying this script.
 
 | parameter | value |
 | --- | --- |
-| code_hash | 0x25c29dc317811a6f6f3985a7a9ebc4838bd388d19d0feeecf0bcd60f6c0975bb |
-| hash_type | type |
+| code_hash | 0x50bd8d6680b8b9cf98b73f3c08faf8b2a21914311954118ad6609be6e78a1b95 |
+| hash_type | data1 |
 | tx_hash | 0xbf6fb538763efec2a70a6a3dcb7242787087e1030c4e7d86585bc63a9d337f5f |
 | index | 0x0 |
 | dep_type | code |
--- a/rfcs/0052-extensible-udt/0052-extensible-udt.md
+++ b/rfcs/0052-extensible-udt/0052-extensible-udt.md
@@ -32,9 +32,9 @@ type:
 lock:
     <user_defined>
 ```
-
-The added `xUDT args` and `xUDT data` parts provide all the new functions needed
-by xUDT, the detailed structure is explained below.
+The `amount` is a 128-bit unsigned integer in little endian format. The added `xUDT
+args` and `xUDT data` parts provide all the new functions needed by xUDT, the
+detailed structure is explained below.
 
 ### **xUDT Args**
 
@@ -44,7 +44,8 @@ xUDT args has the following structure:
 <4-byte xUDT flags> <Variable length bytes, extension data>
 ```
 
-Depending on the content of `flags`, different extension data might be attached:
+Depending on the content of `flags`, which is represented as a 32-bit unsigned
+integer in little-endian format, different extension data might be attached:
 
 • If `flags & 0x1FFFFFFF` is 0, no extension data is required. Note a
 backward-compatible way of viewing things, which is that a plain sUDT cell also
--- a/rfcs/0052-extensible-udt/0052-extensible-udt.md
+++ b/rfcs/0052-extensible-udt/0052-extensible-udt.md
@@ -338,6 +338,16 @@ the spec above has been deployed to Mirana CKB mainnet and Pudge testnet:
 | index | 0x0 |
 | dep_type | code |
 
+This script is not upgradeable due to zero lock (lock args with all zeros).
+We have previously deployed scripts with the ability to be upgraded. However, we
+consider this upgrading mechanism to be harmful as it compromises the
+decentralization of the blockchain. With the ability to upgrade, it means that
+the owners (whether it's several people or even just one person) of this cell
+can upgrade it at any time. This gives one or several individuals complete
+control over the assets associated with the deploy scripts. That's why we have
+chosen to use a zero lock when deploying this script.
+
+
 - Pudge(testnet)
 
 | parameter | value |
--- a/rfcs/0052-extensible-udt/0052-extensible-udt.md
+++ b/rfcs/0052-extensible-udt/0052-extensible-udt.md
@@ -126,7 +126,7 @@ field `raw_extension_data` and `extension_data` are used when `flags &
 
 ### Owner Mode Update
 
-As described in RFC sUDT, If an input cell in the current transaction uses an
+As described in RFC sUDT, if an input cell in the current transaction uses an
 input lock script with the same script hash as the owner lock script hash, the
 `is_owner_mode` will be set to true. In xUDT, this rule is updated by the
 following rule:
--- a/rfcs/0052-extensible-udt/0052-extensible-udt.md
+++ b/rfcs/0052-extensible-udt/0052-extensible-udt.md
@@ -274,7 +274,7 @@ that the hash of raw extension data provide here, is the same as blake160 hash
 included in xUDT args. After this, it uses the same logic as the previous
 workflow.
 
-### Owner Mode Without Consuming Cell
+### Owner Mode without Consuming Cell
 
 As described above, If an input cell uses an input lock script with same script
 hash as the owner lock script hash, the `is_owner_mode` will be set to true. It
--- a/rfcs/0052-extensible-udt/0052-extensible-udt.md
+++ b/rfcs/0052-extensible-udt/0052-extensible-udt.md
@@ -46,7 +46,7 @@ xUDT args has the following structure:
 
 Depending on the content of `flags`, different extension data might be attached:
 
-• If `flags & 0x1FFFFFFF` is 0, we won’t need any extension data. Note a
+• If `flags & 0x1FFFFFFF` is 0, no extension data is required. Note a
 backward-compatible way of viewing things, which is that a plain sUDT cell also
 has a hidden `flags` field with all zeros.
 
--- a/rfcs/0052-extensible-udt/0052-extensible-udt.md
+++ b/rfcs/0052-extensible-udt/0052-extensible-udt.md
@@ -224,7 +224,7 @@ Witnesses:
 
 The witness of the same index as the first input xUDT cell is located by xUDT script. It is parsed first as WitnessArgs structure, the `input_type` or `output_type` field of `WitnessArgs`, is thus treated as `XudtWitnessInput` structure. The length of `extension_data` structure inside must also be the same as `ScriptVec` in `xUDT args`. An extension script might also require transaction-specific data for validation. The witness here provides a place for these data needs.
 
-Notice each extension script is only executed once in the transaction. The extension script is responsible for checking all xUDT cells of the current type, ensuring each cell data and witness for the current extension script, can be validated per the extension script’s rules.
+Note that each extension script is only executed once in the transaction. The extension script is responsible for checking all xUDT cells of the current type, ensuring each cell data and witness for the current extension script, can be validated against the extension script’s rules.
 
 ### **P2SH Style Extension Script**
 
--- a/rfcs/0052-extensible-udt/0052-extensible-udt.md
+++ b/rfcs/0052-extensible-udt/0052-extensible-udt.md
@@ -112,7 +112,7 @@ The `input_type` or `output_type` field in witness has the following data
 structure in molecule format:
 
 ```js
-table XudtWitnessInput {
+table XudtWitness {
     owner_script: ScriptOpt,		
     owner_signature: BytesOpt,
     raw_extension_data: ScriptVecOpt,
@@ -121,8 +121,12 @@ table XudtWitnessInput {
 ```
 
 The field `owner_script` and `owner_signature` will be used in owner mode. The
-field `raw_extension_data` and `extension_data` are used when `flags &
-0x1FFFFFFF` is 0x2 in args.
+field `raw_extension_data` is used when `flags & 0x1FFFFFFF` is 0x2 in args.
+
+The length of `extension_data` structure inside must also be the same as
+`ScriptVec` in `xUDT args` or `raw_extension_data`. An extension script might also
+require transaction-specific data for validation. The witness here provides a
+place for these data needs.
 
 ### Owner Mode Update
 
@@ -145,10 +149,9 @@ won’t be run in a transaction.
 If the `owner_script` in witness isn’t none and its blake2b hash is the same as
 the owner lock script hash in `args`, this script will be run as an extension
 script. If the script returns success, `is_owner_mode` is set to true. Note, the
-`owner_signature` field can be used by this owner script. Usually, a valid
-signature can be placed here. When tokens are minted, the `owner_script` and
-`owner_signature` can be set to some proper values. When tokens are transferred,
-they can be set to none.
+`owner_signature` field can be used by this owner script. When tokens are
+minted, the `owner_script` and `owner_signature` can be set to some proper
+values. When tokens are transferred, they can be set to none.
 
 ### **xUDT Data**
 
@@ -212,7 +215,7 @@ Outputs:
 Witnesses:
     WitnessArgs structure:
         Lock: <user defined>
-        Input Type: <XudtWitnessInput>
+        Input Type: <XudtWitness>
             owner_script: <None>
             owner_signature: <None>				
             raw_extension_data: <None>
@@ -222,9 +225,9 @@ Witnesses:
                 <...>
 ```
 
-The witness of the same index as the first input xUDT cell is located by xUDT script. It is parsed first as WitnessArgs structure, the `input_type` or `output_type` field of `WitnessArgs`, is thus treated as `XudtWitnessInput` structure. The length of `extension_data` structure inside must also be the same as `ScriptVec` in `xUDT args`. An extension script might also require transaction-specific data for validation. The witness here provides a place for these data needs.
+The witness of the same index as the first input xUDT cell is located by xUDT script. It is parsed first as WitnessArgs structure, the `input_type` or `output_type` field of `WitnessArgs`, is thus treated as `XudtWitness` structure.
 
-Note that each extension script is only executed once in the transaction. The extension script is responsible for checking all xUDT cells of the current type, ensuring each cell data and witness for the current extension script, can be validated against the extension script’s rules.
+Note that each extension script is only executed once in the transaction. When multiple instances of the same extension script are included, each instance will execute independently for each inclusion. The extension script is responsible for checking all xUDT cells of the current type, ensuring each cell data and witness for the current extension script, can be validated against the extension script’s rules.
 
 ### **P2SH Style Extension Script**
 
@@ -237,7 +240,7 @@ Inputs:
             <amount: uint128> <xUDT data>
         Type:
             code_hash: xUDT type script
-            args: <owner lock script hash> <xUDT args>
+            args: <owner lock script hash> <xUDT args, hash of raw extension data>
         Lock:
             <user defined>
     <...>
@@ -247,14 +250,14 @@ Outputs:
             <amount: uint128> <xUDT data>
         Type:
             code_hash: xUDT type script
-            args: <owner lock script hash> <xUDT args>
+            args: <owner lock script hash> <xUDT args, hash of raw extension data>
         Lock:
             <user defined>
     <...>
 Witnesses:
     WitnessArgs structure:
         Lock: <user defined>
-        Input Type: XudtWitnessInput
+        Input Type: XudtWitness
             owner_script: <None>
             owner_signature: <None>				
             raw_extension_data: 
@@ -267,7 +270,7 @@ Witnesses:
                 <...>
 ```
 
-The only difference here is that `XudtWitnessInput` in `input_type` or
+The only difference here is that `XudtWitness` in `input_type` or
 `output_type` field in the corresponding WitnessArgs structure, contains raw
 extension data in `ScriptVec` data structure, xUDT script must first validate
 that the hash of raw extension data provide here, is the same as blake160 hash
@@ -301,7 +304,7 @@ Witnesses:
     WitnessArgs structure:
         Lock: <user defined>
         Input Type: <None>
-        Output Type: XudtWitnessInput
+        Output Type: XudtWitness
             owner_script: <owner script 1>
             owner_signature: <signature 1>				
             raw_extension_data: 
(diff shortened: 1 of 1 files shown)


rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,370 @@
...
+
+| parameter | value |
+| --- | --- |
+| code_hash | 0x50bd8d6680b8b9cf98b73f3c08faf8b2a21914311954118ad6609be6e78a1b95 |
+| hash_type | data1 |
+| tx_hash | 0xc07844ce21b38e4b071dd0e1ee3b0e27afd8d7532491327f39b786343f558ab7 |
+| index | 0x0 |
+| dep_type | code |
+
+This script is not upgradeable due to zero lock (lock args with all zeros).
```
Flouse, line 341: When I see the word `zero lock`, I thought it is a lock whose `code_hash` is all zero, not just with all zeros lock args.

## [Some ckb-system-scripts](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0024-ckb-genesis-script-list/0024-ckb-genesis-script-list.md) such as SECP256K1 and Nervos DAO are using lock script
```json
{
  "code_hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
  "hash_type": "data",
  "args": "0x"
}
```

## While xUDT is using
```json
"a_lock_with_zero_args": {
  "code_hash": "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8",
  "hash_type": "type"
  "args": "0x0000000000000000000000000000000000000000",
},
```

I know it is almost impossible to unlock `a_lock_with_zero_args`.
Just got a little confused with the `zero lock` wording.

XuJiandong: It's limited by ckb-cli deployment tools. It can't set code_hash to all zero.

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,370 @@
...
+An xUDT cell is backward compatible with Simple UDT, all the existing rules
+defined in the Simple UDT spec must still hold true for xUDT cells. On top of
+sUDT, xUDT extends a cell like the following:
+
+```yaml
+data:
+    <amount: uint128> <xUDT data>
+type:
+    code_hash: xUDT type script
+    args: <owner lock script hash> <xUDT args>
```
Flouse, line 31:
```suggestion
    code_hash: xUDT type script
    hash_type: data1 on mainnet | type on testnet
    args: <owner lock script hash> <xUDT args>
```

In the current deployments, developers using xUDT will have to switch the `hash_type` of xUDT type_script between testnet and mainnet.

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,370 @@
...
+lock:
+    <user_defined>
+```
+
+The added `xUDT args` and `xUDT data` parts provide all the new functions needed
+by xUDT, we will explain the detailed structure below.
```
Flouse, line 37:
```suggestion
by xUDT, the detailed structure is explained below.
```

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,370 @@
...
+
+xUDT args has the following structure:
+
+```
+<4-byte xUDT flags> <Variable length bytes, extension data>
+```
+
+Depending on the content of `flags`, different extension data might be attached:
+
+• If `flags & 0x1FFFFFFF` is 0, we won’t need any extension data. Note a
```
Flouse, line 49:
```suggestion
• If `flags & 0x1FFFFFFF` is 0, no extension data is required. Note a
```

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,370 @@
...
+    code_hash:      Byte32,
+    hash_type:      byte,
+    args:           Bytes,
+}
+
+vector ScriptVec <Script>
+```
+
+Each entry included in `ScriptVec` structure is interpreted as a CKB script hash
+for an extension script with additional behaviors. When an xUDT script is
```
Flouse, line 68:
```suggestion
Each entry included in `ScriptVec` structure is interpreted as
an extension script with additional behaviors. When an xUDT script is
```

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,370 @@
...
+The field `owner_script` and `owner_signature` will be used in owner mode. The
+field `raw_extension_data` and `extension_data` are used when `flags &
+0x1FFFFFFF` is 0x2 in args.
+
+### Owner Mode Update
+
+As described in RFC sUDT, If an input cell in the current transaction uses an
+input lock script with the same script hash as the owner lock script hash, the
+`is_owner_mode` will be set to true. In xUDT, this rule is updated by the
+following rule:
```
Flouse, line 132:
```suggestion
As described in RFC sUDT, if an input cell in the current transaction has
the same lock script hash with the owner's lock script hash, the
`is_owner_mode` will be set to true. In xUDT, this rule is updated by the
following rule:
```

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,370 @@
...
+    WitnessArgs structure:
+    Lock: <user defined>
+    Input Type: <XudtWitnessInput>
+        owner_script: <None>
+        owner_signature: <None>				
+        raw_extension_data: <None>
+        extension_data: 
+            <vec> BytesVec
+                <data> 
+            <...>
```
Flouse, line 222:
```suggestion
Witnesses:
    WitnessArgs structure:
        Lock: <user defined>
        Input Type: <XudtWitnessInput>
            owner_script: <None>
            owner_signature: <None>				
            raw_extension_data: <None>
            extension_data: 
                <vec> BytesVec
                    <data> 
                <...>
```
add indents

rfcs/0052-extensible-udt/0052-extensible-udt.md
XuJiandong: In spec, we can't write that. We can deploy this script on local devnet with other code_hash/hash_type.

rfcs/0052-extensible-udt/0052-extensible-udt.md
XuJiandong: After updating, it reads: "an input cell has same lock script hash" which is incorrect.

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,370 @@
...
+```
+
+The witness of the same index as the first input xUDT cell is located by xUDT script. It is parsed first as WitnessArgs structure, the `input_type` or `output_type` field of `WitnessArgs`, is thus treated as `XudtWitnessInput` structure. The `extension_data`structure inside must also be of the same length as `xUDT args`. An extension script might also require transaction-specific data so as to validate. Witness here provides a place for this data needs.
```
Flouse, line 225:
```suggestion
The witness of the same index as the first input xUDT cell is located by xUDT script. It is parsed first as WitnessArgs structure, the `input_type` or `output_type` field of `WitnessArgs`, is thus treated as `XudtWitnessInput` structure. The length of `extension_data` structure inside must also be the same as `ScriptVec` in `xUDT args`. An extension script might also require transaction-specific data for validation. The witness here provides a place for these data needs.
```

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,370 @@
...
+
+Notice each extension script is only executed once in the transaction. The extension script is responsible for checking all xUDT cells of the current type, ensuring each cell data and witness for the current extension script, can be validated per the extension script’s rules.
```
Flouse, line 227:
```suggestion
Note that each extension script is only executed once in the transaction. The extension script is responsible for checking all xUDT cells of the current type, ensuring each cell data and witness for the current extension script, can be validated against the extension script’s rules.
```

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,370 @@
...
+extension data in `ScriptVec` data structure, xUDT script must first validate
+that the hash of raw extension data provide here, is the same as blake160 hash
+included in xUDT args. After this, it uses the same logic as the previous
+workflow.
+
+### Owner Mode Without Consuming Cell
+
+As described above, If an input cell uses an input lock script with same script
+hash as the owner lock script hash, the `is_owner_mode` will be set to true. It
+isn’t convenience: this requires extra cell to be consumed. With `owner_script`
```
Flouse, line 281:
```suggestion
isn’t convenient: this requires extra cell to be consumed. With `owner_script`
```

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,370 @@
...
+`output_type` field in the corresponding WitnessArgs structure, contains raw
+extension data in `ScriptVec` data structure, xUDT script must first validate
+that the hash of raw extension data provide here, is the same as blake160 hash
+included in xUDT args. After this, it uses the same logic as the previous
+workflow.
+
+### Owner Mode Without Consuming Cell
+
+As described above, If an input cell uses an input lock script with same script
+hash as the owner lock script hash, the `is_owner_mode` will be set to true. It
```
Flouse, line 280:
```suggestion
As described above, If an input cell in the current transaction has the same
lock script hash as the owner's lock script hash, the `is_owner_mode` will be set to true. It
```

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,370 @@
...
+isn’t convenience: this requires extra cell to be consumed. With `owner_script`
+and `owner_signature` with proper values, we can use owner mode without extra
```
Flouse, line 282:
```suggestion
and `owner_signature` set to proper values, we can use owner mode without extra
```

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,370 @@
...
+```
+
+The only difference here is that `XudtWitnessInput` in `input_type` or
+`output_type` field in the corresponding WitnessArgs structure, contains raw
+extension data in `ScriptVec` data structure, xUDT script must first validate
+that the hash of raw extension data provide here, is the same as blake160 hash
+included in xUDT args. After this, it uses the same logic as the previous
+workflow.
+
+### Owner Mode Without Consuming Cell
```
Flouse, line 277:
```suggestion
### Owner Mode without Consuming Cell
```

rfcs/0052-extensible-udt/0052-extensible-udt.md
Flouse: yep, the original sentence is better.

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,370 @@
...
+}
+```
+
+The field `owner_script` and `owner_signature` will be used in owner mode. The
+field `raw_extension_data` and `extension_data` are used when `flags &
+0x1FFFFFFF` is 0x2 in args.
+
+### Owner Mode Update
+
+As described in RFC sUDT, If an input cell in the current transaction uses an
```
Flouse, line 129:
```suggestion
As described in RFC sUDT, if an input cell in the current transaction uses an
```

doitian (dismissed):

janx (dismissed):

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -167,7 +167,7 @@ table XudtData {
 The `data` field included in `XudtData`, must be of the same length
 as `ScriptVec` structure included in xUDT args. Some extensions might require
 user-specific data stored in each xUDT cell. xUDT data provides a place for such
-data.
+data. This XUDT data part can be optional.
```
roughpandaz, line 170: Nice!

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,369 @@
...
+cells. Under a lot of scenarios, it is `input_type`. But in the following example “Owner
+Mode Without Consuming Cell”, we can see it’s possible on `output_type`.
+
+### xUDT Witness
+
+The `input_type` or `output_type` field in witness has the following data
+structure in molecule format:
+
+```js
+table XudtWitnessInput {
```
xxuejie, line 115: The name here is slightly confusing. While most cases would keep this data in `input_type` field, there are cases where it might be stored in `output_type` field. This means we probably should not name it `XXXInput`.

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,369 @@
...
+- input type script (when `flags & 0x80000000` is **non-zero**)
+
+With the same script hash as the owner lock script hash, the `is_owner_mode`
+will be set to true. The output lock scripts are not included, because they
+won’t be run in a transaction.
+
+If the `owner_script` in witness isn’t none and its blake2b hash is the same as
+the owner lock script hash in `args`, this script will be run as an extension
+script. If the script returns success, `is_owner_mode` is set to true. Note, the
+`owner_signature` field can be used by this owner script. Usually, a valid
```
xxuejie, line 148: Calling this field `xxx_signature`, when the text explicitly mentions this field **usually** contains a signature(hinting there might be case this field does not contain a signature) is so weird.

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,369 @@
...
+table XudtWitnessInput {
+    owner_script: ScriptOpt,		
+    owner_signature: BytesOpt,
+    raw_extension_data: ScriptVecOpt,
+    extension_data: BytesVec,
+}
+```
+
+The field `owner_script` and `owner_signature` will be used in owner mode. The
+field `raw_extension_data` and `extension_data` are used when `flags &
```
xxuejie, line 124: As explained below, `extension_data` is not only used here, it is used in previous cases as well.

xxuejie: And the naming choice here is also misleading. One would think that `raw_extension_data` and `extension_data` are correlated, but in fact they are used in different places.

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,369 @@
...
+            owner_script: <None>
+            owner_signature: <None>				
+            raw_extension_data: <None>
+            extension_data: 
+                <vec> BytesVec
+                    <data> 
+                <...>
+```
+
+The witness of the same index as the first input xUDT cell is located by xUDT script. It is parsed first as WitnessArgs structure, the `input_type` or `output_type` field of `WitnessArgs`, is thus treated as `XudtWitnessInput` structure. The length of `extension_data` structure inside must also be the same as `ScriptVec` in `xUDT args`. An extension script might also require transaction-specific data for validation. The witness here provides a place for these data needs.
```
xxuejie, line 225: I recommend we have a dedicated place explaining the purpose of `extension_data`, instead of hiding it deeply here.

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,369 @@
...
+
+Note that each extension script is only executed once in the transaction. The extension script is responsible for checking all xUDT cells of the current type, ensuring each cell data and witness for the current extension script, can be validated against the extension script’s rules.
```
xxuejie, line 227: > Note that each extension script is only executed once in the transaction.

Is this ensured in the implementation? What if the same extension script is included multiple times on purpose? If so, how?

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,369 @@
...
+When `flags & 0x1FFFFFFF` are set to 0x2, only the blake160 hash of extension data is included in xUDT args. The user is required to provide the actual extension data in witness directly:
+
+```yaml
+Inputs:
+    <vec> xUDT_Cell
+        Data:
+            <amount: uint128> <xUDT data>
+        Type:
+            code_hash: xUDT type script
+            args: <owner lock script hash> <xUDT args>
```
xxuejie, line 240: Better to mark clearly that the `xUDT args` field here contains a hash, not the actual value.

rfcs/0052-extensible-udt/0052-extensible-udt.md
```diff
@@ -0,0 +1,369 @@
...
+                <...>
+            extension_data: 
+                <vec> BytesVec
+                    <data>
+                <...>
+```
+
+The example above shows a scenario of owner mode without consuming the owner's cell. The `<owner lock script
+hash 1>` is the same as blake160 hash of `<owner script 1>`. We can implement an
+extension script as `<owner script 1>` which has same functionality to
```
xxuejie, line 319: Can we rephrase this part? It's so misleading. How can we implement such a thing? Or is it required to implement such a script?

rfcs/0052-extensible-udt/0052-extensible-udt.md
XuJiandong: Updating this name is easy. But the script code is freezed and the script is deployed.
Can we just add some comments about it's possible in `output_type` in some scenarios.

xxuejie: I do understand the code is freezed, I'm not suggesting any change in behaviors.

But I do believe that API-side, we can tweak the name without any breaking code. There is even a chance the binary code would still stay the same even if we modify the field names in molecule source file, recreate molecule generated file, then recompile the C source file, assuming we've done sufficient purging. Since the RFC is still being written, I do recommend we spend enough time thinking about proper names.

rfcs/0052-extensible-udt/0052-extensible-udt.md
XuJiandong: > each extension script is only executed once
It means every extension script is executed one by one, only once. Same extension scripts included multiple times will run for multiple times. There is no optimization.

XuJiandong: When multiple instances of the same extension script are included, each instance will execute independently for each inclusion.

phroi:
Hey Cryptape, iCKB here 👋 I was updating iCKB implementation from sUDT to xUDT and I stumbled upon an apparent inconsistency in the RFC:

- On mainnet xUDT is deployed and referenced by `data1` with a zero lock.
- On testnet xUDT is deployed and referenced by `type`.

So my questions are:

1. Why keeping on testnet the reference by `type` when the implementation on mainnet  cannot be updated? :thinking:
2. Would it be possible to update the proposed RFC to reflect the reference by `data1` for both? If not, why?

Keep up the Great Work,
Phroi

XuJiandong:
> Hey Cryptape, iCKB here 👋 I was updating iCKB implementation from sUDT to xUDT and I stumbled upon an apparent inconsistency in the RFC:
> 
> * On mainnet xUDT is deployed and referenced by `data1` with a zero lock.
> * On testnet xUDT is deployed and referenced by `type`.
> 
> So my questions are:
> 
> 1. Why keeping on testnet the reference by `type` when the implementation on mainnet  cannot be updated? 🤔
> 2. Would it be possible to update the proposed RFC to reflect the reference by `data1` for both? If not, why?
> 
> Keep up the Great Work, Phroi

Done in https://github.com/nervosnetwork/rfcs/pull/428/commits/4c6c76284ca7c94faae23f75be4d1d81ce53e20e

xxuejie:
This part also confuses me: why does the xUDT spec make a choice to use `data1`(or actually any particular value of `hash type`?). To me, it is possible to make recommendations, but it should really be the decision of each individual asset to say if it wants to use `data1` or `type`.

And to me it really is wrong to limit `hash type` to `data1` for testnet deployment to get the same value as mainnet.

I get it that one contributing reason, is that the sole deployment of xUDT on mainnet was done without a type ID, but there is really no stopping for one to deploy an xUDT script by themselves with a type ID setup, then issue a new token using `type` as `hash type`. Are we saying those tokens are not xUDT tokens?

I do recommend that the spec shall not make decisions on the values of `hash type` used by an asset type.

phroi:
Hey Cryptape, iCKB here 👋 I was updating iCKB scripts implementation to xUDT and I found myself in a pickle. 

The specification is not really clear on `xUDT Data`, so I dug a bit the code and I was shocked. While [the molecule file has a definition for `xUDT Data`](https://github.com/nervosnetwork/ckb-production-scripts/blob/master/c/xudt_rce.mol#L67), no validation actually occurs in the xUDT code itself! 🙀 🙀 🙀

This claim is easy to verify as:
- `XudtData` is not referenced in any other molecule definition.
- In all generated validating functions the string `XudtData` always appears in the generated function name.
- The string `XudtData` never appears in the xUDT code itself.

After the initial shock, I understood that it could be a sensible decision as:
- To be 100% sUDT compatible `xUDT Data` rules cannot be possibly enforced in all cases.
- The molecule definition uses `vector Bytes <byte>`, but the actual Bytes should be free from such rules. For example no need for the `Bytes` header if the lock uses a static Struct.
- The `XudtData` rule enforcement is delegated to the Extension Scripts who actually need them for locating their own data.
- This could work similarly to the CoBuild rules enforcement.

@XuJiandong given that now I know the Big Unspoken Secret (🤣), let's talk about `xUDT Data` eye to eye: as it stands now the spec is **not clear** on xUDT data. Developers will naturally interpret it in slightly different ways and and in the future it will become a mess of slightly incompatible implementations. We don't want that! ❌❌❌

How about instead changing the RFC wording to reflect the implementation?

As it stands now:
- If Extension Scripts exist in the xUDT, then the `XudtData` table rules are enforced by the Extension Scripts themselves. Of course you need to **remove the definition of Bytes as vector** as the final data stored by the lock will likely not be a vector.
- If no Extension Scripts exist in the xUDT, **no rules are enforced**. After the fist 16 bytes of UDT amount, all data is managed by the lock and the lock **chooses** how to use it, including the `XudtData` format.

How about changing the spec in this direction?

Keep up the Great Work,
Phroi

phroi:
As far as I can see, the intended a molecule representation of `XudtData` is the following:

```molecule
// Bytes ...                                  // No Bytes definition as it's a placeholder for anything
option BytesOption (Bytes);                   // BytesOption can be of zero length
vector BytesOptionVec <BytesOption>           // This is a dynvec, some fields may have zero length
option BytesOptionVecOption (BytesOptionVec); // BytesOptionVecOption can be of zero length

table XudtDataTable {
  lock: BytesOption,                          // Possibly Empty
  data: BytesOptionVecOption,                 // Possibly Empty
}

option XudtData (XudtDataTable);
```

**Considerations:** in case there are no xUDT Script Extensions and Lock Script stores data as `XudtData`, the only overhead is the table header, so `12 bytes`, is this correct?

@XuJiandong does this represent the intended xUDT behaviour? If so, could you please update the spec to reflect it?

phroi:
@XuJiandong I was wondering: after the 16 bytes for amount and the XudtData table, is it allowed to have additional data? Or it's forbidden and it should be represented inside the XudtData table?

**Example**: a NFT data and metadata could be stored be stored after the XudtData table. Does it makes sense or should be forbidden?

phroi:
Also another question about XudtData design: why was a table of with an inner dynvec was chosen in the first place? Is there any good reason?

It seems like such a waste of precious CKB when the full thing could have been defined as a dynvec and the data of the lock encoded (possibly empty) as first/last position of the dynvec!

The full thing could have had an initial overhead of 8 CKB (with only Script data) and only an additional overhead 4 CKB for every Script Extension, but now with the table & dynvec it's around double that amount 🙀

XuJiandong:
@phroi  All molecule issues can be found in [this RFC](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0008-serialization/0008-serialization.md). Excessive discussion of molecule specifics is not required in this RFC.

phroi:
@XuJiandong molecule spec is fixed and agreed upon, not my concern here. Could you please answer the my questions about xUDT?

XuJiandong:
> Example: a NFT data and metadata could be stored after the XudtData table. Does it makes sense or should be forbidden?
It's not allowed. There's no need to store them after the table; they can safely be placed in the item of data.

phroi:
> they can safely be placed in the item of XudtData.data

Perfect, so we can assume that if there are NFT, there will be a corresponding dedicated extension script 👍 (correct?)


@XuJiandong about the revised molecule representation of XudtData that I proposed earlier (with all those options...), how much I got wrong?

XuJiandong:
> Considerations: in case there are no xUDT Script Extensions and Lock Script stores data as XudtData, the only overhead is the table header, so 12 bytes, is this correct?
@phroi  it's correct

phroi:
Perfect, it wasn't clear for me, thank you!! 🙌 

@XuJiandong it could be a good idea to update the XudtData molecule schema to better represent this optionality 💪

Last question. Given that xUDT is fully compatible with sUDT and that XudtData is optional, in which cases is it ok to not use XudtData representation?

For example, let's assume a xUDT does not use Script Extensions. So normally exactly the 16 bytes of amount are stored in cell data.  No table is used normally. (Correct?)

Then this xUDT is used with a lock that stores data in cell data. From the xUDT standard perspective:

1. This xUDT + lock cell must adopt the XudtData schema.
2. Or it's advisable, but not necessary due to sUDT compatibility.

Which is the correct interpretation from a xUDT standard perspective? 1 or 2?

XuJiandong:
@phroi 
```
option XudtData(XudtDataTable); 
```

You may overlook this definition. While `XudtDataTable` requires at least 16 bytes, `XudtData` is entirely optional (correct my earlier response). In cases where Script Extensions are not used, there may be zero bytes. So choose option 2, leave cell data empty.

If you have any further question, please talk about this issue on discord. This PR is for review only.

XuJiandong:
> The deployment of xUDT on Testnet is TypeId according to the [Cryptape blog](https://blog.cryptape.com/enhance-sudts-programmability-with-xudt#heading-xudt-extension-scripts) and almost Dapps/Wallets/Explorer are using the version with TypeId, see https://pudge.explorer.nervos.org/scripts.
> 
> I'm confused about why the RFC document ultimately chose version data1, given that very few projects in the ecosystem use this version.

This two versions (`data1` and `type1) are pointing to the same version of xUDT. It is a history reason. I will update it in RFC.

xxuejie:
> For example a XUDT referenced by type and one referenced by data1 with the same args are two completely different tokens. For example iCKB xUDT will use the data1.

> Another issue that may arise is that some platforms will support only type, while other only data1 and some both.

Personally, I would disagree with this. To me, it is an unfortunate fact that some platforms only support one value of `hash_type` (typically `type`), but I would not personally bet the design on implementation quirks of specific tools.

The deployment of a script code(whether the cell contains the actual code for a script uses type id or not), is actually decoupled from the fact how another will use such script code. The developer of a CKB script code can deploy their code with type ID enabled, but it is not necessarily the fact that others will have to use the type ID. To me it is perfectly normal to use `data1` or other data variants to reference a script code deployed with type ID, there is nothing wrong with this pattern.

And yes, the same CKB script code referenced using `type` and `data1` will be treated as 2 different scripts, and this is perfectly okay. Personally, I always consider the combination of `code_hash` and `hash_type` to uniquely identify a particular CKB script code. It's really an unfortunate, historical incident, that some platforms only identify a piece of CKB script code via `code_hash`, and just assume certain value of `hash_type`.

phroi:
Since so far seems that we are going for both versions, I asked Lumos if they could integrate the testnet `xUDT` `data1` version, @homura [noticed](https://github.com/ckb-js/lumos/issues/735#issuecomment-2264767948) that at 2024-08-02T04:43:05Z there still was [no transaction record](https://pudge.explorer.nervos.org/script/0x50bd8d6680b8b9cf98b73f3c08faf8b2a21914311954118ad6609be6e78a1b95/data1) for the data1 version of the xUDT script. This seems to represent an issue for the integration in Lumos.

xxuejie (approved):

phroi:
Hooray! 🚀🚀🚀

PR #429, opened 2024-01-19
Update 0051-ckb2023.md

matt-nervos:
Adjust grammar and clarity of language

### DIFF
--- a/rfcs/0051-ckb2023/0051-ckb2023.md
+++ b/rfcs/0051-ckb2023/0051-ckb2023.md
@@ -8,15 +8,15 @@ Created: 2023-04-17
 
 # CKB Consensus Change (Edition CKB2023)
 
-The current edition of CKB consensus rules is CKB2021. CKB2023 refers to the new edition of CKB consensus rules after its second hardfork, The purpose of a hard fork is to upgrade and update the rules encoded in the network. The changes are not backward compatible. This document outlines the changes in this upgrade.
+The current edition of CKB consensus rules is CKB2021. CKB2023 refers to the new edition of CKB consensus rules after the network's second hardfork, the purpose of a hard fork is to upgrade and update the rules encoded in the network. The changes are not backward compatible. This document outlines the changes in this upgrade.
 
 ## What's in CKB2023
-CKB2023 will bring significant changes to the consensus rules, these changes include the removal of the reservation rule on version field in the block header, the introduction of a new version of the virtual machine (VM) with new syscalls and standard extensions, and the optimization of performance with new mops. This RFC provides a detailed overview of these changes.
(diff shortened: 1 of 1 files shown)


rfcs/0051-ckb2023/0051-ckb2023.md
```diff
@@ -26,37 +26,37 @@ Since CKB2023, there will be multiple VM versions available. [RFC32] introduces
...
-* [“A” Standard Extension](https://five-embeddev.com/riscv-isa-manual/latest/a.html), strictly speaking “A” Standard Extension in ckb-vm does not bring functional changes, but many existing code will be compiled with Atomic Instructions and need to be patched, while ckb-vm can implement A instructions to eliminate such work. For example, in CKB VM v2, if you write a script with rust, you can now use [log](https://crates.io/crates/log) crate directly.
-* Introduce more [macro-op fusion](https://en.wikichip.org/wiki/macro-operation_fusion) to reduce cycles consumption of scripts.
+* New syscalls Spawn, Get Memory Limit, Set Content will be added. The syscall Spawn is the core part of this update. Together, the Spawn and the latter two syscalls: Get Memory Limit and Set Content, implement a way to call another CKB Script from within a CKB Script. Unlike the Exec syscall, Spawn saves the execution context of the current script, and, like posix_spawn, the parent script blocks until the child script ends.
```
jordanmack, line 33: > * New syscalls Spawn, Get Memory Limit, Set Content will be added. The syscall Spawn is the core part of this update. Together, the Spawn and the latter two syscalls: Get Memory Limit and Set Content, implement a way to call another CKB Script from within a CKB Script. Unlike the Exec syscall, Spawn saves the execution context of the current script, and, like posix_spawn, the parent script blocks until the child script ends.

Suggestion:

* Three new syscalls will be added: `Spawn`, `Get Memory Limit`, and `Set Content`. These three calls are used together to provide a new method for a script to call another script during execution. The `Spawn` syscall saves the execution context of the currently running script and executes a child script. It blocks until the child script has finished executing, then resumes execution of the parent. This is unlike the currently available Exec syscall, which does not return to the parent, and is very similar to the `posix_spawn` syscall available in POSIX operating systems.
* The RISC-V [A Standard Extension](https://five-embeddev.com/riscv-isa-manual/latest/a.html) will be added to provide atomic instructions in CKB-VM v2. These instructions are highly beneficial operations related to concurrency. Their inclusion makes CKB-VM v2 more compatible with existing Rust crates without the need for patching or other workarounds. For example, in CKB-VM v2, you can now use the [log crate](https://crates.io/crates/log), which was not possible previously without patching.
* Additional [Macro-Operation Fusion](https://en.wikichip.org/wiki/macro-operation_fusion) optimizations will be introduced in CKB-VM v2. This will further reduce the cycle consumption of executing scripts, allowing for more complex functionality on the same hardware.

jordanmack (changes requested):

PR #436, opened 2024-04-19
Update spawn syscalls

mohanson:
PR for ckb: https://github.com/nervosnetwork/ckb/pull/4380

### DIFF
--- a/rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
+++ b/rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
@@ -2,7 +2,7 @@
 Number: "0050"
 Category: Standards Track
 Status: Draft
-Author: Xu Jiandong <lynndon@gmail.com>, Dingwei Zhang <zhangsoledad@gmail.com>
+Author: Xuejie Xiao <xxuejie@gmail.com>, Xu Jiandong <lynndon@gmail.com>, Wanbiao Ye <mohanson@outlook.com>, Dingwei Zhang <zhangsoledad@gmail.com>
 Created: 2023-04-17
 ---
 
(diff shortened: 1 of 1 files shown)


rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -12,174 +12,139 @@ Created: 2023-04-17
...
 
 ```c
-int ckb_get_memory_limit();
+int ckb_inherited_file_descriptors(uint64_t* fd, size_t* count);
 ```
 
-For the script(prime script) directly invoked by CKB, it will always return 8(4M). For the child script invoked by prime script or other child script, it depends on the parameters set by *Spawn*.
+When returning from the syscall, the syscall fills `fd` with the file descriptors in unit of `uint64_t` and fills in `count` with the count of corresponding file descriptors. The actual count of file descriptor written to `fd` is the minimum value between the count of `inherited_fds` in the Spawn syscall and the input value pointed to by `count` before syscall.
+
+By providing `NULL` as `fd` and a `size_t` pointer with a value of 0 as `count`, this syscall can be used to fetch the count of `inherited_fds` in the Spawn syscall. However, by providing `NULL` as `fd` and a non-zero value as `count`, this syscall fails.
```
xxuejie, line 95: Why is it a failure to write `fd` to `NULL`? What if I simply want to stores all my fds starting from address 0? Historically we never forbids this behavior, I'm not sure if there is any gains.

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -12,174 +12,139 @@ Created: 2023-04-17
...
 
-### Current Memory
-[Current Memory]: #current-memory
+### Read
+[Read]: #read
 
-Get the Current Memory Usage. The result is the sum of the memory usage of the parent script and the child script.
+This syscall reads data from a pipe via a file descriptor. The syscall Read attempts to read up to value pointed by length bytes from file descriptor fd into the buffer, and the actual length of data read is written back to the length parameter. The syscall may pause the execution of current process.
```
xxuejie, line 100: I think we will need a more detailed description of read & write syscalls:

* What could contribute to the fact that less than requested bytes are read / written?
* Will the VM block so read syscalls can get more bytes to read? Will the VM block so write syscalls can write more bytes?
* And it would be best if we can describe a way to tell a channel is closed by using read / write syscall

I do recommend checking out the man page for Linux read / write syscalls.

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -12,174 +12,139 @@ Created: 2023-04-17
...
 
-- Length up to 256K.
-- If the written length is greater than the limit given by *Spawn*, the final written length is the minimum of the two.
-- Returns 0 for success, otherwise for failure.
-- This function is optional. Not every child script needs to call this, and nothing will happen if you call it from the prime script.
+### Close
+[Close]: #close
 
-### Spawn example
+This syscall manually closes a file descriptor. After calling this, any attempt to read/write the file descriptor pointed to the other end would fail.
```
xxuejie, line 118: Like mentioned above, please describe in details how read/write to a closed channel would fail. This can be a very useful utility.

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -12,174 +12,139 @@ Created: 2023-04-17
...
+It's not always necessary to manually close file descriptors. When a process is terminated, all file descriptors owned by the process are automatically closed.
 
-```c
-#include <stdint.h>
-#include <string.h>
+### Wait
+[Wait]: #wait
 
-#include "ckb_syscalls.h"
+The syscall pauses until the execution of a process specified by `pid` has ended.
```
xxuejie, line 129: What if parent A spawns child B, but child B terminates before parent A calls `ckb_wait`? I know you do have an answer, but I believe we should properly define the behavior here.

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -12,174 +12,139 @@ Created: 2023-04-17
...
+    - 1: read from witness
+- bounds: high 32 bits means offset, low 32 bits means length. if length equals to zero, it read to end instead of reading 0 bytes.
+- spawn_args: pass data during process creation or save return data.
+    - argc: argc contains the number of arguments passed to the program
+    - argv: argv is a one-dimensional array of strings
+    - process_id: a pointer used to save the process_id of the child process
+    - inherited_fds: an array representing the file descriptors passed to the child process. It must end with zero.
 
-If the sub-script does not write data, then `NULL` can be passed in for `content` and `content_length`.
+The arguments used here index, source, bounds, place, argc and argv follow the usage described in [EXEC].
```
xxuejie, line 71: I don't think the grammar here is correct, please fix it.

xxuejie (approved):

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
mohanson: I removed the special handling of NULL

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
mohanson: Add more detailed under the Write system call:

- If the Writer writes W bytes and the Reader reads R bytes where W > R, the Writer will block, and the Reader will return immediately with R bytes in `*length`. The Reader can then call `ckb_read` again to read the remaining W - R bytes.
- If the Writer writes W bytes and the Reader reads R bytes where W <= R, both the Writer and the Reader will return immediately with W bytes in `*length`.

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -12,174 +12,147 @@ Created: 2023-04-17
...
 ```
 
-We can call this dependent library in the prime script. The prime script passes in two parameters "hello", "world" and checks if the return value is equal to "helloworld”:
+It's not always necessary to manually close file descriptors. When a process is terminated, all file descriptors owned by the process are automatically closed.
+
+### Wait
+[Wait]: #wait
 
-**prime.c**
+The syscall pauses until the execution of a process specified by `pid` has ended. If the process has already terminated when Wait is called, the method returns immediately.
```
xxuejie, line 137: The method returns immediately, but with what return code? In addition, will it update `exit_code` part?

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -12,174 +12,147 @@ Created: 2023-04-17
...
 
-- Length up to 256K.
-- If the written length is greater than the limit given by *Spawn*, the final written length is the minimum of the two.
-- Returns 0 for success, otherwise for failure.
-- This function is optional. Not every child script needs to call this, and nothing will happen if you call it from the prime script.
+Read and Write must appear in pairs; otherwise, the process will be blocked indefinitely. When using `ckb_read` and `ckb_write`, there may be the following scenarios:
 
-### Spawn example
+- If the Writer writes W bytes and the Reader reads R bytes where W > R, the Writer will block, and the Reader will return immediately with R bytes in `*length`. The Reader can then call `ckb_read` again to read the remaining W - R bytes.
+- If the Writer writes W bytes and the Reader reads R bytes where W <= R, both the Writer and the Reader will return immediately with W bytes in `*length`.
```
xxuejie, line 116: I do recommend that we include more cases here, for example: if a VM tries to read from a channel, but no other VM writes to that channel, will the VM reading data hang there? Or will there be early return in some cases? We should really cover all those cases.

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -12,174 +12,147 @@ Created: 2023-04-17
...
 
-Consider the creation of a dependency library with a straightforward function that receives parameters, concatenates them, and subsequently returns the resulting string to the caller.
+### Close
+[Close]: #close
 
-**lib_strcat.c**
+This syscall manually closes a file descriptor. After calling this, any attempt to read/write the file descriptor pointed to the other end would fail. After using close, there are four typical situations:
```
xxuejie, line 121: It's worth mentioning that closing a single file descriptor, essentially closes the entire pair of channels.

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -213,5 +186,240 @@ This syscall might return the following errors:
...
+
+- Error code 5: The file descriptor is invalid during syscall [Wait].
+- Error code 6: The file descriptor is not owned by this process.
+- Error code 7: The other end of the pipe is closed.
+- Error code 8: The maximum count of spawned processes has been reached.
+- Error code 9: The maximum count of created pipes has been reached.
+
+It's possible for read/write/wait operations to wait for each other, leading to a deadlock state. In such cases, CKB-VM would throw an internal error and terminate immediately.
+
+## Cycles
```
xxuejie, line 201: When a VM gets suspended (and maybe later resumed), will it consume cycles? Can we document them here as well?

xxuejie (commented):
There are some hard limits to the system, such as:

* How many VMs can be instantiated at the same time?
* The maximum number of VMs(including suspended ones) a scheduler can spawn?

Are those documented anywhere?

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
mohanson: Added two scenarios about hang. Scenarios about writing or reading from a closed fd are introduced in ckb_close.

mohanson:
> There are some hard limits to the system, such as:
> 
> * How many VMs can be instantiated at the same time?
> * The maximum number of VMs(including suspended ones) a scheduler can spawn?
> 
> Are those documented anywhere?

Descriptions have been added about the number of vm, the number of instantied vm and the number of pipes.

15168316096 (approved):
The scenarios [0, 1, 1], [1, 0, 1], and [1, 1, 0] have already been verified.

XuJiandong (dismissed):

quake (approved):

XuJiandong (approved):

PR #439, opened 2024-05-08
Add details to supported BTC addresses in omnilock

XuJiandong:


quake (approved):

Hanssen0:
Omnilock has not fully supported the P2SH BTC address format, and we can not distinguish P2SH-P2WPKH addresses from other P2SH addresses. Considering this, mentioning P2SH-P2WPKH addresses without warning can cause permanent locking of assets.

We should remove the P2SH-P2WPKH address format or at least explicitly warn developers.

XuJiandong:
I think it can be mentioned that the omnilock only supports P2SH-P2WPKH only but not generic P2SH.

Hanssen0:
Do you think we need to tell developers that we strongly recommend NOT to use P2SH addresses to avoid mis-locking?

janx (approved):

xxuejie (approved):

PR #445, opened 2024-10-31
Noting that process scheduling in ckb-vm is deterministic

mohanson:
The developer may suspect that there will be randomness in the implementation of spawn, so this sentence is added intentionally.

### DIFF
--- a/rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
+++ b/rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
@@ -228,7 +228,7 @@ Here are the main scenarios that may lead to deadlock:
 0. **Circular Waiting**: If multiple processes are in the `Wait`, `WaitForWrite`, or `WaitForRead` states and are waiting on each other in a circular dependency, a deadlock can occur. For example, if:
     - Process A is in `WaitForRead` for data from Process B
     - Process B is in `WaitForRead` for Process A. Both processes will wait indefinitely, as each is waiting for the other to proceed.
-0. **Buffer Limits**: If processes rely on pipes with limited buffer sizes and one process blocks on a `WaitForWrite` state because the pipe buffer is full, and the reader process is also blocked in a `WaitForRead` state (but on a different file descriptor), this can create a deadlock if neither can proceed:
+0. **Buffer Limits**: The pipe in ckb-vm is unbuffered. If one process blocks on a `WaitForWrite` state because the data is not fully read, and the reader process is also blocked in a `WaitForRead` state (but on a different file descriptor), this can create a deadlock if neither can proceed:
     - Process A wants to read 10 bytes from fd0, and then read 10 bytes from fd1, and finally read 10 bytes from fd0.
     - Process B writes 20 bytes into fd0, and then write 10 bytes into fd1.
 
--- a/rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
+++ b/rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
@@ -213,7 +213,25 @@ Five new error types added:
 - Error code 8: The maximum count of spawned processes has been reached.
 - Error code 9: The maximum count of created pipes has been reached.
 
-It's possible for read/write/wait operations to wait for each other, leading to a deadlock state. In such cases, CKB-VM would throw an internal error and terminate immediately.
+## Deadlock
+
+Deadlock is a situation where two or more processes are unable to proceed because they are each waiting for resources or conditions that can only be provided by another waiting process. In the context of this scheduler, where processes communicate via pipes and can enter various states, such as `Runnable`, `Wait`, `WaitForWrite`, `WaitForRead`, `Terminated`. In our scheduler, deadlock will occur if all unterminated processes are waiting and no process is in a runnable state.
+
+- The process enters the `Wait` state by calling the `wait()`
+- The process enters the `WaitForWrite` state by calling the `write()`
+- The process enters the `WaitForRead` state by calling the `read()`
+- The process enters the `Runnable` when a process is created, or it get returned from `wait()`, `write()` and `read()`.
+- The process enters the `Terminated` when a process is terminated.
+
+Here are the main scenarios that may lead to deadlock:
+
+0. **Circular Waiting**: If multiple processes are in the `Wait`, `WaitForWrite`, or `WaitForRead` states and are waiting on each other in a circular dependency, a deadlock can occur. For example, if:
+    - Process A is in `WaitForRead` for data from Process B
+    - Process B is in `WaitForRead` for Process A. Both processes will wait indefinitely, as each is waiting for the other to proceed.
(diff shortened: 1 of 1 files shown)


mohanson:
cc @xxuejie @XuJiandong

xxuejie:
I recommmend expanding this section a bit:

* For each hardfork version, the process scheduling will be deterministic, any indeterminism will be treated as critical / security bugs that requires immediate intervention
* However, based on real usage on chain, it is expected that future hardfork versions would improve the process scheduling workflow, hence making the behavior different across versions

Simply put, the scheduler included in CKB will be deterministic with respect to a particular hardfork version, and will likely change its behavior in different hardfork versions.

XuJiandong:
Provide detailed information about the deterministic scheduler, including:
1. The criteria for selecting the next process to run when one process is blocked.
2. The conditions under which deadlock can occur.
3. The specific operations that can cause a process to become blocked.

xxuejie:
> Provide detailed information about the deterministic scheduler, including:
> 
> 1. The criteria for selecting the next process to run when one process is blocked.
> 2. The conditions under which deadlock can occur.
> 3. The specific operations that can cause a process to become blocked.

I'm not entirely sure here: are those information suitable for RFC, or separate documentations?

An analogy here, could be the details of the transaction pool, those are also highly tied to actual implementation, yet the RFCs does not fully describe those.

mohanson:
I suggest putting the detailed working principle of the scheduler in a separate document.

xxuejie:
A second thought: I think it makes sense to document the deadlock conditions for each hardfork versions here; but other details fit better in a separate document on Nervos docs.

mohanson:
> A second thought: I think it makes sense to document the deadlock conditions for each hardfork versions here; but other details fit better in a separate document on Nervos docs.

There is already descriptions of when read/write/wait will cause deadlock 

https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md#write
https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md#error-code

xxuejie:
Yes but the words are quite vague:

> If ckb-vm detects that all processes are blocked, ckb-vm will return a deadlock error.
> It's possible for read/write/wait operations to wait for each other, leading to a deadlock state.

I think it makes sense to specifically define when all process will be blocked, and what it means for operations to wait for each other.

XuJiandong (approved):

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -208,7 +213,26 @@ Five new error types added:
 - Error code 8: The maximum count of spawned processes has been reached.
 - Error code 9: The maximum count of created pipes has been reached.
 
-It's possible for read/write/wait operations to wait for each other, leading to a deadlock state. In such cases, CKB-VM would throw an internal error and terminate immediately.
+## Deadlock
+
+Deadlock is a situation where two or more processes are unable to proceed because they are each waiting for resources or conditions that can only be provided by another waiting process. In the context of this scheduler, where processes communicate via pipes and can enter various states, such as `Runnable`, `Running`, `Terminated`, `WaitForExit`, `WaitForRead`, `WaitForWrite`. In our scheduler, deadlock will occur if all unterminated processes are waiting and no process is in a runnable state.
+
+- The process enters the `Runnable` when a process is created, or it get returned from `wait()`, `write()` and `read()`.
+- The process enters the `Running` when a process is running.
```
xxuejie, line 221: when a process starts running

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -208,7 +213,26 @@ Five new error types added:
 - Error code 8: The maximum count of spawned processes has been reached.
 - Error code 9: The maximum count of created pipes has been reached.
 
-It's possible for read/write/wait operations to wait for each other, leading to a deadlock state. In such cases, CKB-VM would throw an internal error and terminate immediately.
+## Deadlock
+
+Deadlock is a situation where two or more processes are unable to proceed because they are each waiting for resources or conditions that can only be provided by another waiting process. In the context of this scheduler, where processes communicate via pipes and can enter various states, such as `Runnable`, `Running`, `Terminated`, `WaitForExit`, `WaitForRead`, `WaitForWrite`. In our scheduler, deadlock will occur if all unterminated processes are waiting and no process is in a runnable state.
+
+- The process enters the `Runnable` when a process is created, or it get returned from `wait()`, `write()` and `read()`.
```
xxuejie, line 220: I would personal rephrase here: it's not that a particular syscall returned, it's really that the blocking condition for a process is resolved, so the process enters runnable state.

If still confused, I recommend looking into literature for operating systems.

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -208,7 +213,26 @@ Five new error types added:
...
+- The process enters the `Running` when a process is running.
+- The process enters the `Terminated` when a process is terminated.
+- The process enters the `WaitForExit` state by calling the `wait()`
```
xxuejie, line 223: Simply calling `wait` won't necessarily trigger the process to enter `WaitForExit` state, it really is that when process a calls wait on process b, but process b is still running. In other words, process a now has a blocking condition.

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -208,7 +213,26 @@ Five new error types added:
...
+- The process enters the `WaitForRead` state by calling the `read()`
```
xxuejie, line 224: Similar to the above, a process might not actually enter `WaitForRead` state by calling `read`, if data are already available at the other end. It only enters this state when it wants data but data are not ready, in other words, it has a blocking condition.

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -208,7 +213,26 @@ Five new error types added:
...
+- The process enters the `WaitForWrite` state by calling the `write()`
```
xxuejie, line 225: Similar to the above

rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md
```diff
@@ -208,7 +213,26 @@ Five new error types added:
...
+
+Here are the main scenarios that may lead to deadlock:
+
+0. **Circular Waiting**: If multiple processes are in the `Wait`, `WaitForWrite`, or `WaitForRead` states and are waiting on each other in a circular dependency, a deadlock can occur. For example, if:
+    - Process A is in `WaitForRead` for data from Process B
+    - Process B is in `WaitForRead` for data from Process A. Both processes will wait indefinitely, as each is waiting for the other to proceed.
+0. **Buffer Limits**: Essentially, it's another circular waiting. The pipe in ckb-vm is unbuffered. If one process blocks on a `WaitForWrite` state because the data is not fully read, and the reader process is also blocked in a `WaitForRead` state (but on a different file descriptor), this can create a deadlock if neither can proceed:
```
xxuejie, line 232: It really depends on perspective, to me, this is another example of circular dependency.

xxuejie (approved):

XuJiandong (approved):

PR #448, opened 2025-03-18
Arbitrary multisig, a multisig design for CKB which supports multiple digital signature algorithms

xxuejie:
[PREVIEW LINK](https://github.com/xxuejie/rfcs/blob/multisig/rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md)


rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md
```diff
@@ -0,0 +1,354 @@
...
+
+Unlike `multisig configuration`, the `unlocking configuration` will be frequently seen on chain, most likely in one of the witnesses, to provide data for a multisig validation flow to succeed.
+
+An unlocking configuration is defined as follows:
+
+```
+S ++ R ++ M ++ N ++ (UNLOCKING_SLOT ...)
+```
+
+An unlocking configuration uses the same `S`, `R`, `M`, `N` value as the corresponding multisig configuration. The unlocking configration uses `UNLOCKING_SLOT`s to provide exactly `M` valid signatures for the multisig validation flow to verify. In addition, the first `R` `UNLOCKING_SLOT`s must each have a signature attached.
```
XuJiandong: typo `configration`

rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md
```diff
@@ -0,0 +1,354 @@
...
+* 1 encoded in [VLQ](https://en.wikipedia.org/wiki/Variable-length_quantity) format.
+* 20 encoded in [VLQ](https://en.wikipedia.org/wiki/Variable-length_quantity) format.
+* witness source in 64-bit little-endian unsigned integer
+* witness index in 32-bit little-endian unsigned integer
+* witness offset in 32-bit little-endian unsigned integer
+* witness length in 32-bit little-endian unsigned integer
+
+Note that pipe is capable of accepting binary data with zeros, the data above do not need zero escape encoding.
+
+The 4 variables (witness source / index / offset / length) included in the above signature data packet, jointly locates a slice of data in a particular witness field from the current CKB transction. The leaf script then treats the located slice of data as a series of `UNLOCKING_SLOT` structure, then performs the following verification:
```
XuJiandong: typo `transction`

rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md
```diff
@@ -0,0 +1,354 @@
...
+
+* If any `UNLOCKING_SLOT` uses a value of `ALOG_ID` which is not supported by the current leaf script, the leaf script halts.
```
XuJiandong, line 259: typo `ALOG_ID`

rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md
```diff
@@ -0,0 +1,354 @@
...
+
+* 0 encoded in [VLQ](https://en.wikipedia.org/wiki/Variable-length_quantity) format.
+* 0 encoded in [VLQ](https://en.wikipedia.org/wiki/Variable-length_quantity) format.
+* 0 encoded in [VLQ](https://en.wikipedia.org/wiki/Variable-length_quantity) format.
+
+The leaf script then waits to read more packets for verification from `root_to_leaf_pipe[0]`. There is no situation when a leaf script terminates with a success exit code. We rely on CKB's behavior that when the root VM terminates, all the child VMs also terminate.
+
+### Signature Packet Format 2
+
+Signature Packat Format 1 contains the following data concatenated together:
```
XuJiandong: typo `Packat`.

rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md
```diff
@@ -0,0 +1,354 @@
...
+
+We do believe the arbitrary multisig can be an important building block for future CKB applications.
+
+## Scope
+
+It is worth pointing out that the arbitrary multisig design discussed here is merely a workflow for validating multi-signatures. While a multisig lock script might mostly just contain the arbitrary multisig design in its validation workflow, the arbitrary multisig design is architected in a way that it can be used in more cases than lock scripts. Some type scripts might also leverage the arbitrary multisig design for collective voting or more use cases.
+
+# Terminology
+
+In this document we will have certain psuedocode included, the following operators might be used by psuedocode:
```
XuJiandong: typo `psuedocode`: should be `pseudocode`

rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md
```diff
@@ -0,0 +1,354 @@
...
+
+## `UNLOCKING_SLOT`
+
+An `UNLOCKING_SLOT` resembles `IDENTITY`, but provides an optional signature as well:
+
+```
+(ALGO_FLAG in VLQ encoding) ++ (Public key, public key hash or more composite data) ++ (An optional signature)
+```
+
+In other words,  typical `UNLOCKING_SLOT` contains `ALGO_FLAG`, a public key or public key hash, and an optional signature. Unlike `IDENTITY`, the `SIGNATURE_FLAG` in `UNLOCKING_SLOT` can be set or cleared, denoting the presence or absense of a signature at the end. `ALGO_ID` tells the correct lengths of public key, public key hash, and the optional signature. `ALGO_FLAG` in a whole determines the total length of the `UNLOCKING_SLOT`.
```
XuJiandong: typo `absense`

rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md
```diff
@@ -0,0 +1,354 @@
...
+
+When expected, a root script invokes spawn syscall to create a new VM instance running the leaf script. The root script passes the following data to the leaf script. All the data listed here are concatenated together, encoded via `zero escaped encoding` discussed above, then be put in `argv[0]`:
+
+* The letter `s` in 1-byte ASCII encoding
+* The length of signing message, in 32-bit little-endian unsigned integer
+* The actual signing message in variable length
+
+One can deduce that one spawned leaf script will always run signature verifications using the same signing message.
+
+Before executing spawn syscall, the root script must create 2 pipes via [pipe](https://github.com/nervosnetwork/rfcs/blob/bd5d3ff73969bdd2571f804260a538781b45e996/rfcs/0050-vm-syscalls-3/0050-vm-syscalls-3.md#pipe) syscalls. There will be a totally of 4 file descriptors created from 2 pipes. We will name them differently:
```
XuJiandong: `There will be a total of 4 file`

rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md
```diff
@@ -0,0 +1,354 @@
...
+
+For `ALGO_ID` from `48` to `59`, SPHINCS+ public key for each parameter set will follow the `ALGO_ID`. Each parameter set might have a different length depending on the SPHINCS+ public key following the 1-byte `ALGO_ID`.
+
+We shall come back to define the `IDENTITY` for `ALGO_ID` from `60` to `62` at a later stage.
+
+For the `ALGO_ID` of `63`, the public key part of an `IDENTITY`, contains more data than just the public key, the following concatenated structure is used:
+
+* 32-byte value denoting `code hash`
+* 1-byte value denoting `hash type`
+* 8-byte value denoting `bounds`
```
XuJiandong, line 103: suggest splitting into 2 parts explicitly: length and offset in VQL format. Normally there are only 2 bytes of zeros.

rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md
```diff
@@ -0,0 +1,354 @@
...
+* 0 encoded in [VLQ](https://en.wikipedia.org/wiki/Variable-length_quantity) format.
+* 0 encoded in [VLQ](https://en.wikipedia.org/wiki/Variable-length_quantity) format.
+
+The leaf script then waits to read more packets for verification from `root_to_leaf_pipe[0]`. There is no situation when a leaf script terminates with a success exit code. We rely on CKB's behavior that when the root VM terminates, all the child VMs also terminate.
+
+This signature packet format is in fact designed for `ALGO_ID` of `63`, when an arbitrary script is loaded for signature verification. Signature packet format 2 can be used to exchange signatures to valid between root script and the arbitrary leaf script.
+
+## Exec
+
+A root script might use exec syscall to invoke a leaf script, when the following conditions are met:
```
XuJiandong, line 334: If there is a security issue or malicious code in the leaf script executed via `exec`, an early `exit(0)` can bypass all validation. It should be used with extreme caution.

For example, Alice, Bob, and Charlie agree to use a 2-of-3 signature scheme, but Charlie is aware of a bug in a leaf script that allows `exit(0)` to bypass validation. As a result, he can exploit it to proceed with only one signature.

xxuejie: For one thing, a multisig implementation is fine that it just support spawn syscalls.

To really discuss the code here, I think this is too strong a convention:

```
Charlie is aware of a bug in a leaf script that allows exit(0) to bypass validation
```

Exactly how will this bug proceed? 

To me, it could also happens that Charlie is aware of a bug in a root script that turns out to be `exit(0)`. In this case Charlie just exploits the root script. The fact that the return code of a leaf script becomes the return code of the root script, does not always mean one should always assume the worst.

What if I assume the following:

```
Charlie finds a way that a leaf script sends all zeros via the response pipe.
```

How can we tell one is easier to achieve than the other?

rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md
xxuejie: I think having `bounds` as full 8 bytes actually simply implementation. If you look at most of the code now, a script is fine ignoring VLQ in almost all cases. But making `bounds` 2 VLQ means one will definitely have to implement VLQ, which is another burden. Maybe one day VLQ will grow popularity, but for now, I consider it a spec that lacks precise definition, and want to just restrict its usage a bit.

XuJiandong: I notice that VLQ is used in the following `The length of true public key`.

rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md
XuJiandong: Maybe we could include a chapter on security, explaining various scenarios.

rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md
xxuejie: Yes, and for most public keys, they are well under 127 bytes, which fit a single byte of VLQ, meaning the leaf script actually do not need to implement VLQ.

An ELF binary will most likely be larger than 127 bytes, if `bounds` is expressed in `offset` and `length`, the `length` field will require a VLQ implementation.

Look across the spec, you could see this principle applied in many places: VLQ is only used in 2 places:

* Making sure request / response are compatible with ckb-script-ipc
* Places where in most cases a value will be less than 127

rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md
xxuejie: Done

rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md
```diff
@@ -0,0 +1,376 @@
...
+
+A multisig configuration is defined as follows:
+
+```
+S ++ R ++ M ++ N ++ (IDENTITY ...)
+```
+
+The multisig configuration designed here, is highly inspired(though differences still exist) from [multisig lock included in the genesis block](https://github.com/nervosnetwork/ckb-system-scripts/blob/934166406fafb33e299f5688a904cadb99b7d518/c/secp256k1_blake160_multisig_all.c#L17-L28):
+
+* `S`: a 1-byte reserved value, right now `S` must be `0x80`, to differentiate from the reserved value used in multisig lock included in the genesis block.
```
Hanssen0, line 135: > I insist that we should at least explain this clearly in the document. From the last script to now, the only thing that is available in the public documentation is a description of the reserved area, but there is no description of its function or why we chose the value that way. This has caused confusion.

For full context see https://github.com/cryptape/quantum-resistant-lock-script/pull/14

xxuejie: Just to fill in the complete context: in https://github.com/cryptape/quantum-resistant-lock-script/pull/14#discussion_r2137122749, @Hanssen0 actually proposed a different design: we can simply remove `S`, whether it is a reserved field or a version field, a lock / type script implementing current multisig design should implement this multisig design only. When a new multisig design is introduced, a new script should be used, different `code_hash` shall be used to distinguish one multisig design from another.

Not to say which direction is better, but just want to provide a complete background here, we can then discuss which design adapts to the future better.

xxuejie: Adding a counter example: in the design of cobuild, we actually encoded special union IDs:

```
union WitnessLayout {
  SighashAll: 4278190081,
  SighashAllOnly: 4278190082,
  Otx: 4278190083,
  OtxStart: 4278190084,
}
```

See [here](https://talk.nervos.org/t/ckb-transaction-cobuild-protocol-overview/7702/9) for more details.

Those special union IDs, are exactly hacks required, because `WitnessArgs` do not have a version byte, or a reserved field. In future expansions, we will need to tell the 2 formats apart(`WitnessArgs` vs `cobuild data structures`). Even if an old script only deals with `WitnessArgs`, it still brings merits since it can reject apps that wrongly use the new cobuild format.

So personally, I do see values in a reserved field in the multisig format. Even if we want to stick to the fact that one script only implements a single multisig format, it still can benefit us in the future.

phroi:
There is an incoherence in this RFC implemented by the quantum resistant lock: 
1. `S` needs to be **explicit** ([`version field`](https://github.com/cryptape/quantum-resistant-lock-script/pull/14#discussion_r2137055860)).
2. Set of supported digital signature schemes (by a given lock implementing the RFC such a the current) is **implicit** and requires off-chain knowledge to interact with said lock ([`though only a subset of ALGO_IDs is supported`](https://github.com/xxuejie/rfcs/blob/fb238472852f12e16221900b38297183e614083a/rfcs/0000-arbitrary-multisig/0000-arbitrary-multisig.md#examples)).

Do we need off-chain knowledge about the lock we are interacting with?

- If **no**, ok, prove it: design a test suite for the RFC that doesn't need to be tailored for a given lock implementing this RFC (no hand picked exclusion of tests depending on the lock, that's cheating)
- If **yes**, we can drop this field, as it is not necessary.

phroi:
Additionally, **where are the examples/visualizations in this RFC?**

The section named `examples` is just a low effort link to an incomplete implementation of this RFC in the form of the quantum resistant lock. I'd call it low effort because it's very much unclear what's the mapping between the entities described in the RFC and the ones appearing in the code.

Are we even trying to get other devs to understand the work being done here?

[Please provide examples and visualizations](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0017-tx-valid-since/0017-tx-valid-since.md)!! :pray:

PR #452, opened 2026-04-08
Remove the A extension from the ckb-vm 2023 hardfork description file

mohanson:
Ref https://github.com/nervosnetwork/ckb-vm/pull/510


XuJiandong:
syscall spawn should be updated, too.

Matt-RUN-CKB:
is there any write-up about why the A extension wasn't included in ckb2023?

XuJiandong:
This can be resolved at the library level: https://github.com/nervosnetwork/ckb-std/blob/master/src/dummy_atomic.rs

Since this was removed before the hard fork, it should not appear in the RFC. The RFC was apparently not updated to reflect this change.

zhangsoledad (approved):

XuJiandong (approved):

PR #454, opened 2026-05-20
Update link in DAO withdrawal process gotchas

eval-exec:
Add PR link for the "64 output cells limitation": https://github.com/nervosnetwork/ckb-system-scripts/pull/93

@phroi @chenyukang @zhangsoledad @quake


phroi:
@eval-exec you see, that's the issue: most auditors still pick-up the design without the 64 bit limitation, which is **not** the one deployed on-chain.

Looking at the commit timestamps, the version actually committed on mainnet should be:

https://github.com/nervosnetwork/ckb-system-scripts/blob/f25c5ae8824c4907ad94326a0113a03defab9bfc/c/dao.c 

The commit message reads "fix: NervosDAO review fixes" and it's dated 2019/11/02

The one instead I would  like to propose as link in the DAO RFC is dated 2020/05/25, it has the same code except it greatly improve comments:

https://github.com/nervosnetwork/ckb-system-scripts/blob/814eb82c44f560dbdad2be97eb85464062920237/c/dao.c

Also it was the last commit on dao.c before my relatively recent PR.

Phroi

